PR #62, opened 2015-01-23
Calculate acquisition statistics

jeanconn:
Add code to calculate per-obsid per-star acquisition success and write out a data table.  This new table includes details such as error flags and acquisition attempts that are not available in the sybase acquisition database.  Other useful columns such as CCD temperature at acquisition and expected warm pixel fraction are included.  

Todo:

- [x] Documentation
- [x] Interface code or guidelines


mica/stats/acq_stats.py
```diff
@@ -0,0 +1,572 @@
...
+               second_guide['AOATTQT2'],
+               second_guide['AOATTQT3'],
+               second_guide['AOATTQT4']])
+    dq = q2 / q1
+    dot_q = np.sum(q1.q * q2.q)
+    if dot_q > 1:
+        dot_q = 1.0
+    return dq, dot_q
+
+
```
jeanconn: I think the broadcasting wasn't correct for this application in mica.quaternion.  This should eventually get pushed back there if it is the *right thing*.

taldcroft:
@jeanconn - please always provide a useful description when creating a pull request, esp. a big one.

jeanconn:
Todo:

- [ ] Documentation
- [ ] Interface code or guidelines

taldcroft:
- Is this a strict superset of the acq data in `star_stat_db`?
- Will / should `acq_stat_report`  be updated to use this?

jeanconn:
I'm continually improving matching the right catalogs to the star data, so this may not be a strict superset of star_stat_db.

I don't see a downside in updating acq_stat_report to use this table (when actually installed).

I think we'll also want to write out the table as sqlite to make it easier to eventually integrate with starcheck (unless we've done a full python transition before we get there).

taldcroft:
This `acq_stats` table is one of the more useful new things we have on the burner.  What about those docs and getting this merged?  :smile:

jeanconn:
I think these docs are passable for this for the time being.  Any other concerns before this gets merged?

taldcroft:
Can you put in units where applicable?

jeanconn:
I thought I had, but I guess all those angle ones probably could use some
arcsecs...

taldcroft:
Generally speaking it is easier to see the units if you put them at the end in parens, or even better square brackets.  In some cases you have the units "in-lined" like "one shot delta quaternion arc length in arcsecs for this obsid manvr", but that's tough to see.

doc/acquisition_data.rst
```diff
@@ -22,195 +20,82 @@ and for each acquisition star determines
...
+  from mica.stats.acq_stats import get_stats
+  acq_data = get_stats()
+
+Alternatively, the raw hdf5 may be read directly.  It includes the following columns:
+
+
+=============== ====================================================================
+ Column         Description
+=============== ====================================================================
+obsid           the obsid
```
taldcroft: Don't need "the"

jeanconn: Agreed, though I didn't touch that for these changes.

taldcroft:
:+1:

PR #63, opened 2015-02-28
WIP: Add a generic pcad table app

jeanconn:
Adds a more general purpose pcad table display.  This takes a time range and displays content similar to the pcad acq table display.

### DIFF
--- a/mica/web/close_events.py
+++ b/mica/web/close_events.py
@@ -1,0 +1,24 @@
+from kadi import events
+from Chandra.Time import DateTime
+
+# 30 minute pad for these events
+interest_pad = 30 * 60;
+interest = [events.normal_suns,
+            events.safe_suns,
+            events.scs107s,
+            events.caps]
+
+def close_events(start, stop):
+    es = []
+    for etype in interest:
+        e = etype.filter(start=DateTime(start).secs - interest_pad,
+                         stop=DateTime(stop).secs + interest_pad)
+        if e.count():
+            es.append(e)
+    me = events.major_events.filter(DateTime(start) - 1,
+                                    DateTime(stop) + 1)
+    if me.count():
+        es.extend(me)
+    return es
+
+
--- a/mica/web/templates/mica/telem_table.html
+++ b/mica/web/templates/mica/telem_table.html
@@ -1,0 +1,159 @@
+{% extends "base.html" %}
+
+{% block extra_css %}
+<style>
+table.pcad td {
+line-height: 1rem;
+padding: .4rem;
+font-size: .7rem;
+}
+table.pcad tr {
+background-color: transparent;
+}
+table.pcad {border-spacing: 5px;}
+</style>
+{% endblock %}
+
+{% block body %}
+<body class="with-background">
+{% endblock %}
+
+{% block content %}
+<p><p>
+<h1 align="center"> PCAD Telem Table </h1>
+
+<form action="" method="get">
+      <div class="row">
+        <div class="large-2 columns"> &nbsp; </div>
+        <div class="large-3 columns">
+          <label for="right-label1" class="right inline">Start Time</label>
+          <label for="right-label2" class="right inline">Stop Time</label>
+        </div>
+        <div class="large-3 columns">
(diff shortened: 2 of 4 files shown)


mica/web/templates/mica/telem_table.html
```diff
@@ -0,0 +1,159 @@
...
+        </div>
+        <div class="large-4 columns"> &nbsp; </div>
+      </div>
+      <input type="submit" value="Submit"  style="display: none;">
+</form>
+
+{% if events %}
+<TABLE class='pcad'>
+{% for event in events %}
+<TR><TD>{{ event }}</TD></TR>
```
jeanconn: This could use clever formatting to get clean descriptions and links back to event details, but I didn't see already-available code to do that for any kind of event.

mica/web/close_events.py
```diff
@@ -0,0 +1,24 @@
...
+from Chandra.Time import DateTime
+
+# 30 minute pad for these events
+interest_pad = 30 * 60;
+interest = [events.normal_suns,
+            events.safe_suns,
+            events.scs107s,
+            events.caps]
+
+def close_events(start, stop):
```
jeanconn: My thought here was that most of the time if I'm interested in "what happened around this time" it would be handy to know if any already-known interesting events occurred close to the time of interest without having to do a separate search in another page.

mica/web/templates/mica/telem_table.html
```diff
@@ -0,0 +1,159 @@
...
+{% endfor %}
+</TABLE>
+{% endif %}
```
jeanconn: This event table should also higher on the page and share vertical space with the time range selector.

mica/web/views.py
```diff
@@ -54,3 +56,48 @@ def get_context_data(self, **kwargs):
...
+            try:
+                start_time = DateTime(start_time)
+                start_time.date
+            except:
+                start_time = None
+                context['start_time_date'] = ''
+        if stop_time is not None:
+            try:
+                stop_time = DateTime(stop_time)
+                stop_time.date
```
jeanconn: I think I threw this in because I wasn't getting an error in the "try" block on an empty time (just got the error later when trying to use it).

mica/web/close_events.py
```diff
@@ -0,0 +1,24 @@
+from kadi import events
+from Chandra.Time import DateTime
+
+# 30 minute pad for these events
+interest_pad = 30 * 60;
+interest = [events.normal_suns,
```
taldcroft: This could be a list of tuples `(event_type, start_pad, stop_pad)`, then you don't need to special-case major_events below.  The asymmetric pad allows normal_sun, safe_suns to have pads of (e.g.) -1 hr to +3 days.  That represents the typical impact window for those events.

mica/web/close_events.py
```diff
@@ -0,0 +1,24 @@
+from kadi import events
+from Chandra.Time import DateTime
+
+# 30 minute pad for these events
+interest_pad = 30 * 60;
```
taldcroft: Module-level constants should be upper case.

mica/web/close_events.py
taldcroft: This probably belongs as a utility in kadi.events.  Which of course makes development trickier, though you can toss it in the kadi web branch.

taldcroft:
I think the `telem_table.py` module should be `pcad_table.py` since it is still fairly specific to PCAD.

jeanconn:
I just had different lists of msids for each so broke it out so there
wouldn't be constant/column collision.  ​

javierggt:
Yes, clearly there is no fire under it... 7 years! 😄 

As it stands, I would say this will never get merged. Especially because from now on this will live in kadi-apps, plus the whole html template is different in kadi-apps. So... if there is anything valuable here, it can be distilled into an issue in kadi-apps, and this should really be closed.

jeanconn:
Though the acquisition table idea upon which this is based should continue to work on the new server, so seems like this path to make a new more generic table could still work if we ever want to do it.

PR #65, opened 2015-06-11
aca_dark update broken as of pyyaks 3.3.3

jeanconn:
```
File "/proj/sot/ska/arch/x86_64-linux_CentOS-6/lib/python2.7/site-packages/mica/archive/aca_dark/update_aca_dark.py", line 38, in <module>
    MICA_FILES = pyyaks.context.ContextDict('update_mica_files', basedir='/tmp')
  File "/proj/sot/ska/arch/x86_64-linux_CentOS-6/lib/python2.7/site-packages/pyyaks-3.3.3-py2.7.egg/pyyaks/context.py", line 251, in __new__
    "({} vs. {})".format(name, basedir, CONTEXT[name].basedir))
ValueError: Re-using context name 'update_mica_files' but basedirs don't match (/tmp vs. /data/aca/archive)
```

### DIFF
--- a/mica/archive/aca_dark/update_aca_dark.py
+++ b/mica/archive/aca_dark/update_aca_dark.py
@@ -25,17 +25,13 @@ from . import dark_cal
 logger = None
 ZODI_PROPS = None
 
-DARK_CAL = pyyaks.context.ContextDict('dark_cal')
+# Bring in definitions from dark_cal pyyaks
+DARK_CAL = dark_cal.DARK_CAL
 
-# Pointer to Ska processing files.  This is hardwired to flight Ska since dark
-# cal files are an OTS input here.
-SKA_FILES = pyyaks.context.ContextDict('ska_files', basedir='/proj/sot/ska')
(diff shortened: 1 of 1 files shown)


taldcroft:
By design it uses a global registry so different modules can share a file structure.  So a line like:
```
SKA_FILES = pyyaks.context.ContextDict('ska_files', basedir='/proj/sot/ska')
```
is not a redefinition, it is a re-use of the same context object (see https://github.com/sot/pyyaks/blob/master/pyyaks/context.py#L247).  But if you change the directory then it is not a re-use.

This is a little like the python logger that has a global registry of logger names.

jeanconn:
OK.  I saw it as both redefinition and re-use, but I take your point.  Does my commit, then, overtrim in the interest of making it easier for me to read?

taldcroft:
Your commit is good and improves overall code clarity.  The real use-case for re-use is where you can't or don't want to import a different module with the definition.

jeanconn:
I tested this with a local MICA_ARCHIVE with no errors.

PR #66, opened 2015-06-30
Improve V&V logging

jeanconn:
Improve V&V logging by reducing verbosity, adding higher level notes (which obsid, which aspect interval), and adding a warning for multiple aspect intervals.


taldcroft:
I'm not planning on a detailed code review but :+1: on the ideas.

PR #67, opened 2015-07-15
Only rerun on new data

jeanconn:
This code was attempting to get new asp1 products and rerun V&V
for all obsids that had "pending" or provisional data.  These changes
set it to fetch and check the version numbers of the content it has
and the "default" content and try to only fetch and redo V&V
(or mark to have V&V redone) if the state in the database has really
changed.

### DIFF
--- a/mica/archive/obsid_archive.py
+++ b/mica/archive/obsid_archive.py
@@ -328,7 +328,7 @@ class ObsArchive:
                 "select distinct obi from obidet_0_5 where obsid = %d" % obsid)
         if len(obis) > 1:
             minobi = np.min(obis['obi'])
-            logger.info("limiting arc5gl to obi %d" % minobi)
+            logger.debug("limiting arc5gl to obi %d" % minobi)
             arc5.sendline("obi=%d" % minobi)
         if version != 'default':
             arc5.sendline("version=%s" % version)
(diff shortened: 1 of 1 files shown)


mica/archive/obsid_archive.py
```diff
@@ -753,7 +766,15 @@ def update(self):
                         current_status['quality'] = 'X'
                 else:
                     # if there are no science_2 entries at all for the obsid
-                    # try an update anyway
+                    # see if it is discarded
+                    with Ska.DBI.DBI(dbi='sybase', server='sqlsao', database='axafocat') as db:
+                        target_status = db.fetchone(
+                            "select status from target where obsid = {}".format(obs['obsid']))
+                        if target_status['status'] == 'discarded':
+                            logger.info("Skipping {}, obsid 'discarded'".format(
```
jeanconn: Also, skip obsids that are now 'discarded' as they are never going to have 'default'/released products in the archive.

taldcroft:
What's your test process for a change like this?

jeanconn:
> What's your test process for a change like this?
>

Yeah, it stinks.   I need to both write a proper test mode and write
something to create a more complete test environment including the pieces I
need about status from the real archive and the ocat and apstat tables.

I tested this somewhat on the side with MICA_ARCHIVE.  Then, I tested the
info-only routines by just calling them on the real archive (unit test
inspection?) and then ran the bulk of these changes with with breakpoints
set before the pieces that do real updates to our tables.   And then I just
tested it live with the knowledge that I could just rebuild the asp1
archive if I blew it up.

jeanconn:
Do you want to hold more logic updates until a better test framework
is in place, or just run with this?

taldcroft:
Your testing sounds adequate, so run with this.

PR #68, opened 2015-09-10
Add function to get obspar obsids within a date range

jeanconn:
There doesn't seem to be a convenient way to get all the obsids that have an available obspar file in the archive within a `start` / `stop` date range.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -323,6 +323,12 @@ Methods are also provided to retrieve a list of files by obsid and time range.
    ...                          stop='2012:030')
 
 
+And a method is provided to fetch obsids that have obspars:
+
+   >>> obspar.get_obsids('2016:001', '2016:002')
+   [51365, 18736, 18036, 18203]
+
+
 .. _stats-label:
 
 Mica.stats
--- a/mica/archive/obspar.py
+++ b/mica/archive/obspar.py
@@ -344,11 +344,14 @@ def get_obsids(start=None, stop=None, revision=None):
       >>> from mica.archive import obspar
       >>> obsids = obspar.get_obsids('2015:001', '2015:003')
 
+    To retrieve the complete set of obsids used over a time range,
+    use kadi (kadi.events.obsid) instead of this obspar-based method.
 
     :param start: time range start (Chandra.Time compatible)
     :param stop: time range stop (Chandra.Time compatible)
-    :param revision: revision integer or 'last'
-                     defaults to current released version
+    :param revision: None, revision integer, or 'last'
+                     None is equivalent to limiting the search
+                     to those with 'default'/released obspars.
     :returns: list of obsids
     """
     files = archive._get_file_records(start=start, stop=stop,
--- a/mica/archive/obspar.py
+++ b/mica/archive/obspar.py
@@ -337,6 +337,27 @@ def get_files(obsid=None, start=None, stop=None, revision=None):
                              revision=revision)
 
 
+def get_obsids(start=None, stop=None, revision=None):
+    """
+    List obsids with obspars in a time range.
+
+      >>> from mica.archive import obspar
+      >>> obsids = obspar.get_obsids('2015:001', '2015:003')
+
+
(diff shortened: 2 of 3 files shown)


jeanconn:
Code attached.

taldcroft:
Looks OK.  Only niggles are:

- `archive._get_file_records()` is looking like a public method now that it gets used from another module.
- It's never too late to start adding unit testing.  :smile:

jeanconn:
Thought I'm wondering now if it makes more sense to just implement this as a public method in obsid_archive instead of obspar.

taldcroft:
OK in principle, but I'm a little confused about obsids vs. obspars.  My understanding is that there are obsids with no obspar, so it's confusing to me if something like `obsid_archive.get_obsids()` only returns obsids with an obspar.  But maybe mica only has archive info for obsids with an obspar?  In any case the method name `get_file_records` should be made less generic.

jeanconn:
If you just want the obsids in a range, don't you want eng_archive / kadi?  I think of mica.archive as our area of easy access to our copy of CXCDS processed products.  mica.archive.obsid_archive has the interface to grabbing and processing those products that are associated with / fetched via obsid.  How would you change the name of get_file_records?

jeanconn:
obsid_archive._get_file_records will currently return the information of any asp1 file type or obspar in the collection with optional filter on CONTENT type and revision.

taldcroft:
I guess you just need to add / work on the doc strings.  I think within you mica work you have internalized "obsid_*" to mean "* that fundamentally maps onto a single ObsId".  This applies to words like ObsArchive or "obsid telemetry" or "obsid_archive".  But to an outsider those mean "an archive of obsids" or "telemetry containing the obsid".  So when I'm looking at the code, even the existing docstring for `ObsidArchive`, I don't quite get what it's doing.  And the double meaning of `archive` (CXCDS archive vs. mica archive) doesn't help.

jeanconn:
I don't disagree, but I think the language in the top level docs isn't
horrible for this.  Also, do you have ideas for this double-meaning of
"archive".  I called it mica.archive because that's what you called out in
your original spec and I think it was just L1 files in that original spec.
 ​

taldcroft:
If you mean the module docstring, then yes that is clear.  But the `ObsArchive` doc string needs clarification and the other methods need to have docstrings.

jeanconn:
I meant the RST / Sphinx docs, but they (module docs, class docstrings,
method docstrings) could all use more help / time / effort.

jeanconn:
Looking over lingering PRs...  From your earlier comment:

> My understanding is that there are obsids with no obspar, so it's confusing to me if something like obsid_archive.get_obsids() only returns obsids with an obspar. But maybe mica only has archive info for obsids with an obspar?

obsid_archive is in the API docs, but it is basically a private module designed to aggregate the functions for the two "per-obsid" collections in the archive?   Did you find a reference to use it from something or were you just seeing what existed under mica.archive?  Should I hide obsid_archive with a "_"?  

I agree that obsid_archive.get_obsids(), would be confusing, but it also doesn't exist nor is it made by this code.  This code would make archive.obspar.get_obsids() (right?).  I'm thinking it might be even more clear to have it make archive.obspar.get_obsids_with_obspars().

taldcroft:
Anyway, after all that I think this is good to merge as-is.  I recently ran into a desire to have all the obsids with obspars in a time range, so let's just do it.  And of course kadi is the right way to find all the obsids in a time range.

jeanconn:
I don't even know what I meant by passing the revision keyword to this wrapper though.  Though as a user, would you want an obsid that doesn't have a "default" obspar yet (still not released in V&V or broken), or would you want all of the obsids that have any kind of obspar?

taldcroft:
I think in this case I would like "clean" obspars that are through V&V and archived.  But it isn't a strong preference, so if that's a pain don't worry.

jeanconn:
That's easy as coded.  I'll set that to be the default and update the
docstring.

PR #70, opened 2015-09-11
Change path to icxc for vv login

jeanconn:
icxc2 is now deprecated for vv login.  This changes the path back to icxc.


taldcroft:
:+1:

PR #72, opened 2015-12-24
Add parameter to specify scaling factor in dark_temp_scale

taldcroft:
This changes the parametrization so that `scale_4c` is the multiplicative *increase* for each 4 degC *increase*.  Thus a bigger value is more increase.  That makes more sense and matches the way it was done for annealing.

This routine is back-compatible however.

### DIFF
--- a/mica/archive/aca_dark/dark_cal.py
+++ b/mica/archive/aca_dark/dark_cal.py
@@ -44,25 +44,31 @@ def dark_id_to_date(dark_id):
     return '{}:{}'.format(dark_id[:4], dark_id[4:])
 
 
-def dark_temp_scale(t_ccd, t_ccd_ref=-19.0):
+def dark_temp_scale(t_ccd, t_ccd_ref=-19.0, scale_4c=1.0 / 0.70):
     """
     Return the multiplicative scale factor to convert a CCD dark map from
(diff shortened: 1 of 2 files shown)


taldcroft:
Tests pass (in particular test_dark_temp_scale).

taldcroft:
Actually I want to do this slightly differently, do not merge.

jeanconn:
I still wonder if we might want a wrapper for dark_temp_scale that has better doc strings about going *from* temperature 1 *to* temperature 2.  I have to really stare at the docstring for the method now to remember it is giving me the scale to get back to the "reference" temp from the non-reference temp.

taldcroft:
Yes, I have the same problem.  I'll improve the docstring now.

jeanconn:
If you don't want to wrap the whole thing, maybe an example with t_ccd_ref not at -19?  Would that help us out next time?

taldcroft:
Does the new docstring work for you?

jeanconn:
Would this be more clear or more confusing if we set it to take a variable number of arguments?  (convert from t_ccd1 to t_ccd2 if given, else use t_ccd_ref)  Or just fixed the few bits of old code so they don't use the t_ccd_ref=-19 default?  Not sure if it is worth the effort, but floating the ideas anyway.

taldcroft:
This is good enough IMHO.

PR #74, opened 2016-01-15
Deprecate local mica versions of star probability functions 

taldcroft:
This is some untested code that I found lying in a branch on disk.  It's probably a good idea for cleanup.

### DIFF
--- a/mica/archive/aca_dark/dark_model.py
+++ b/mica/archive/aca_dark/dark_model.py
@@ -2,12 +2,12 @@
 Routines related to the dark current model and guide / acq success prediction.
 """
 
-from itertools import izip
 import numpy as np
 from numpy import exp, log, arange
 
 import Ska.Numpy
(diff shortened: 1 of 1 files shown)


jeanconn:
I think this is fine, but I still have an action to review the docs and see if we've got any of the deprecated calls as examples (unlikely).

taldcroft:
This is actually done already by different means.

jeanconn:
Can you remind me about the means?

taldcroft:
Well `mica.archive.aca_dark.dark_model` now consists of just 3 lines:
```
import warnings
warnings.warn("mica.archive.aca_dark.dark_model is deprecated.  See chandra_aca.dark_model", DeprecationWarning)
from chandra_aca.dark_model import *
```
So all of the code in this PR is irrelevant.  If there was code that was looking to mica for calculating star probabilities then it would have broken long ago.

PR #76, opened 2016-01-16
Set starcheck parser to ignore new P_ACQ column

jeanconn:
Set starcheck parser to ignore new P_ACQ column


jeanconn:
This parsing code needs modernization but this is a quick fix to keep the parsing working once the starcheck catalog output is changed to include the P_ACQ column in starcheck 11.6.

jeanconn:
Tested with a testing MICA_ARCHIVE and a version each of the 11.5 and 11.6dev starcheck outputs.

PR #79, opened 2016-05-09
Add star history kadi app

jeanconn:
Add star history kadi app

This is intended to replace

https://icxc.harvard.edu/cgi-bin/aspect/get_stats/get_stats.cgi?id=118494392

### DIFF
--- a/mica/web/star_hist.py
+++ b/mica/web/star_hist.py
@@ -38,7 +38,8 @@ def get_gui_data(agasc_id):
     :returns: list of dicts of uses as guide stars
     """
     with DBI(dbi='sybase', server='sybase', user='aca_read') as db:
-        gui = db.fetchall('select * from trak_stats_data where id = {}'.format(
+        gui = db.fetchall(
+            'select * from trak_stats_data where type != "FID" and id = {}'.format(
                 agasc_id))
     if not len(gui):
         return []
--- a/mica/web/star_hist.py
+++ b/mica/web/star_hist.py
@@ -4,6 +4,13 @@ from Ska.DBI import DBI
 
 
 def get_acq_data(agasc_id):
+    """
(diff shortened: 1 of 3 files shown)


mica/web/star_hist.py
```diff
@@ -0,0 +1,51 @@
+from astropy.table import Table
+from mica.stats import acq_stats
+from Ska.DBI import DBI
+
+ACQ_COLS = ['date', 'type', 'obsid', 'obi', 'slot', 'mag',
+            'mag_obs', 'id', 'stat_flags', 'star_tracked', 'acq_dy', 'acq_dz']
+
+GUI_COLS = ['date', 'type', 'obsid', 'obi', 'slot', 'mag',
+            'mag_obs', 'perc_not_track']
```
jeanconn: This column defs should be cut as they aren't used.

taldcroft:
Some description on what this is providing as well as function docstrings and code comments would be good.

jeanconn:
I probably don't need the star_hist code at all, but my templating isn't great, so just making lists of dicts of what I want is a cheap cheat.

jeanconn:
It would be better for me to store the column order and formatting in this code instead of in the template, but this is probably fine for a version 1.

jeanconn:
I should add time range filtering; that will let me use the app to cover and display whatever range we want in starcheck.

jeanconn:
Should also add some notes to the page describing what some of the more ambiguous acquisition columns (star_tracked, dy, dz) are.

taldcroft:
Looks good to me!

taldcroft:
This particular code doesn't yet need to run or be tested against Py3, but it's worth checking that it is at least syntax compatible with a spin through 2to3.

PR #82, opened 2016-05-15
Handle case of completely missing track mag in star report

jeanconn:
Handle case of completely missing track mag in star report


jeanconn:
This allows the reports for obsid 18826 and 51015 to complete.  Each have a sub-report for a star that was used once as a guide star and has no mag history.    The mean of the empty set was making that part of the report fail.

taldcroft:
:+1:

PR #81, opened 2016-05-17
Improve multiple aspect interval warnings

jeanconn:
For warnings like:

"Warning: obsid 18842 has 2 aspect intervals"

it would be good to included "(has monitor window)" if, in fact, the observation does.  If it doesn't an email warn on this error would be useful.


jeanconn:
The unit tests don't exist yet to easily check this, but I can verify that obsid 18842 completes V&V without error or warning, and obsid 14457 gets the appropriate warning: "Warning: obsid 14557 has 4 aspect intervals".

taldcroft:
:+1:

PR #85, opened 2016-08-25
Changes to get obsid 18625 to display in kadi acq table app

jeanconn:
Changes to get obsid 18625 to display in kadi acq table app

This also has the side-effect that the display will include the times through the start of KALM if present, and 5 minutes of data if there is no transition to KALM in the interval.

### DIFF
--- a/mica/version.py
+++ b/mica/version.py
@@ -14,7 +14,7 @@ NOTE: this code copied from astropy and modified.  Any license restrictions
 therein are applicable.
 """
 
-version = '0.5.4dev'
+version = '0.5.4'
 
 _versplit = version.replace('dev', '').split('.')
 major = int(_versplit[0])
--- a/mica/web/pcad_table.py
+++ b/mica/web/pcad_table.py
@@ -85,7 +85,9 @@ def get_acq_table(obsid):
     # Get the catalog for the stars
     # This is used both to map ACQID to the right slot and
     # to get the star positions to estimate deltas later
-    starcheck = mica.starcheck.get_starcheck_catalog(int(obsid))
+    timeline_at_acq = mica.starcheck.starcheck.get_timeline_at_date(manvr.start)
+    starcheck = mica.starcheck.get_starcheck_catalog(int(obsid),
+                                                     mp_dir=timeline_at_acq['mp_dir'])
     if 'cat' not in starcheck:
         raise ValueError('No starcheck catalog found for {}'.format(obsid))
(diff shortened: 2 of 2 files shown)


taldcroft:
I don't know the code well enough to really review.  I just noticed style comments which are optional:
- In general don't break lines at << 80 characters (i.e. line break before `mpdir=mpdir`)
- The Python ternary works and is recommended.  In this case I usually use extra parentheses for readability):
```
mp_dir = None if (timeline_at_acq is None) else timeline_at_acq['mp_dir']
```

jeanconn:
Do you do anything besides try to read off your column position from the status line in emacs to work with 80 characters?  I think I've just been working with a large font and have a bad eye for this.

taldcroft:
I use pyflakes and have it warn for lines > 100.  That is my personal limit.

jeanconn:
Hi @taldcroft .  I've forgotten, why do we have mica in '/proj/web-kadi/lib/python2.7/site-packages' again?  I just updated the version in skare but I think we're getting the older version in kadi web.

taldcroft:
I think that was a leftover from the time of separate web branches.  For that matter I see `kadi` there as well.  Given that kadi no longer has a web branch, it seems both of those should be removed from the local `site-packages`.  I already moved `mica` and confirm that the 18625 displays properly.

Do you agree I should move the local `kadi` away (into `HIDE` for safekeeping)?

jeanconn:
That sounds reasonable, but I don't know what we'd need to confirm to guarantee it is the right thing.  Just a test on the web that the applications seem to show the right content would probably be enough.  I also noted that I think your local python directory shows up in the kadi search path.  Not sure about that either.

jeanconn:
>  that I think your local python directory shows up in the kadi search path

Looks like that is gone now, so I'm not sure which part was bringing it in.

taldcroft:
Humm, doing `grep web-kadi httpd.conf` shows a bunch of stuff, so it would *unwise* to just move that away without thinking.  I think the answer at the moment is that updates to kadi which impact the web need to be installed to /proj/web-kadi following the documented process.

jeanconn:
Though it seems to me if you can directly edit the kadi files called by the `httpd.conf`, you might as well have edit rights on the `httpd.conf` too.  But I don't recall how much of this was outlined in your original setup document for the server.  It looks to me like the documented process needs a tweak (to remove mica if it works just fine from HEAD skare) and maybe a clarification of "as needed" for kadi.

PR #86, opened 2016-09-02
Update CDA script and improve debugging output

jeanconn:
Update CDA script and improve debugging output.

Current failures in the cron job aren't reproducible at the cmd line and there isn't enough output from the cron job failures to see what is going on.  This PR should at least close the hdf5 file properly and might give enough debugging output to see  if there is a problem with the source data from the url.

### DIFF
--- a/mica/archive/cda_aca0_list.py
+++ b/mica/archive/cda_aca0_list.py
@@ -145,15 +145,17 @@ def update_cda_table(data_root=None,
                 row['filetime'] = file['filetime']
                 row.append()
             tbl.flush()
+    # Just fetching errors instead of quitting with bad status
+    except urllib.error.URLError as err:
+        logger.info(err)
+    # Otherwise print the new lines if available to log and reraise
     except Exception as err:
         if new_lines is not None:
             logger.info("Text of attempted table")
             logger.info("".join(new_lines))
-            raise(err)
+        raise(err)
     finally:
         h5f.close()
-
-
 
 def main():
     opt = get_options()
--- a/mica/archive/cda_aca0_list.py
+++ b/mica/archive/cda_aca0_list.py
@@ -1,4 +1,21 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+Maintain a table with a list of of aca level 0 files in the Chandra Data Archive.
+
+This module calls the CGI available from:
+
+https://icxc.harvard.edu/dbtm/CDA/aspect_fetch.html
+
+fetches the list of aca0 files, and updates a full list of files in 'MICA_ARCHIVE/aca0/cda_aca0.h5'.
+That list should be of all available aca level 0 files in the Chandra Data Archive.  When called
+from the update_aca_l0.py update script in the mica cron job, this module attempts to use the end
+of the complete data in cda_aca0.h5 to form the query for new files, so only recent file names
+are fetched.
+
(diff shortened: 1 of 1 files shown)


jeanconn:
I think I can run this on the side for a bit and confirm that the table looks the same as the official table.

mica/archive/cda_aca0_list.py
```diff
@@ -50,11 +49,11 @@ def make_data_table(lines):
     filetimes = [int(file_re.search(f).group(1)) for f in files['filename']]
     versions = [int(file_re.search(f).group(2)) for f in files['filename']]
     now_dates = np.repeat(DateTime().date, len(ingest_dates))
-    files = Ska.Numpy.add_column(files, 'ingest_date', ingest_dates)
-    files = Ska.Numpy.add_column(files, 'aca_ingest', now_dates)
-    files = Ska.Numpy.add_column(files, 'filetime', filetimes)
-    files = Ska.Numpy.add_column(files, 'version', versions)
-    files.sort(order=['aca_ingest', 'ingest_date', 'filename'])
+    files.add_column(Column(ingest_dates, 'ingest_date'))
```
taldcroft: `files['ingest_date'] = ingest_dates` is the recommended idiom now.

taldcroft:
While you are at it, a module docstring saying what this does would be good.  I don't know with certainty what CDA even stands for (Chandra Data Archive would be my guess), and I guess this relates to some custom web query script that arcops made for you?

jeanconn:
> While you are at it, a module docstring saying what this does would be
> good. I don't know with certainty what CDA even stands for (Chandra Data
> Archive would be my guess), and I guess this relates to some custom web
> query script that arcops made for you?
>
​I don't know if I'd call it "custom".  it was intended to be a generic
interface to get the list of Chandra Data Archive files of various types by
date so that we could then do arc45gl queries by filename or some such:

​https://icxc.harvard.edu/dbtm/CDA/aspect_fetch.html

I've been using the aca0 list of files to see if we've got everything in
the mica version of the archive.

mica/archive/cda_aca0_list.py
```diff
@@ -36,8 +35,8 @@ def get_options():
 
 
 def make_data_table(lines):
-    files = asciitable.read(lines,
-                            names=['filename', 'status', 'ingest_time'])
+    files = Table.read(lines, format='ascii.csv',
```
jeanconn: Add guess=false per @taldcroft

jeanconn:
It looks like the astropy.table changes I made aren't completely compatible with the manual pytables hdf5 row appending this code was doing.  @taldcroft do you have any example code using the astropy.table hdf5 support to add rows to an hdf5 table?  Without rewriting the whole thing?

taldcroft:
Astropy `Table` does not have direct backend support for hdf5 row manipulation.  In general Table serves just as a container and not an active database proxy.  Blaze might do this, but it's an heroic undertaking.  Sorry.

jeanconn:
Good to know.  Thanks @taldcroft

jeanconn:
It looks like the change in  d12e9c6 was really all I needed to make this work again.  Lately this seems to be timing out when run as a cron task, but I'm not seeing that when running manually.  This change should just log the URL or HTTP errors instead of exiting and also includes other minor fixes and debugging content.

taldcroft:
I don't think I can meaningfully comment on this now, so don't wait for me.

PR #88, opened 2016-10-11
Add details about brightest pixel in search box to acq data [Currently for Analysis: Do not merge]

jeanconn:
Add details about brightest pixel in search box to acq data.  This uses the dark cal image before the observation as the reference for that bright pixel.

### DIFF
--- a/mica/stats/acq_stats.py
+++ b/mica/stats/acq_stats.py
@@ -476,7 +476,7 @@ def calc_stats(obsid):
         # input for get_starcheck_catalog
         if obsid in [1966]:
             starcheck = get_starcheck_catalog(obsid,
-                                              mp_dir='/2002/JAN1202/oflsa')
+                                              mp_dir='/2002/JAN1202/oflsa/')
         elif obsid in [3105, 2741, 61334, 61333, 61332, 61331, 3358, 3357]:
             starcheck = get_starcheck_catalog(obsid,
                                               mp_dir='/2002/JAN2802/oflsd/')
--- a/mica/stats/acq_stats.py
+++ b/mica/stats/acq_stats.py
@@ -14,6 +14,7 @@ import Ska.quatutil
 import Ska.astro
 from mica.quaternion import Quat
 import mica.archive.aca_dark.dark_model
+from chandra_aca import yagzag_to_pixels
 import logging
 
 import warnings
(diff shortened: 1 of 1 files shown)


mica/stats/acq_stats.py
```diff
@@ -542,17 +548,45 @@ def calc_stats(obsid):
...
-
-
-def table_acq_stats(obsid_info, acq_stats, star_info, catalog, temp):
+    return obsid_info, acq_stats, star_info, catalog, temps, dc_max
+
+
+def get_dark_map_bright(time, catalog):
+    dark_cal = mica.archive.aca_dark.get_dark_cal_props(time, select='before', include_image=True)
+    image = dark_cal['image']
+    row_idx = np.repeat(np.arange(-512, 512), 1024).reshape(1024,1024)
```
jeanconn: This could use a comment to indicate that each of these is a "number matrix" to check and assist with indexing.

mica/stats/acq_stats.py
```diff
@@ -542,17 +548,45 @@ def calc_stats(obsid):
...
+    col_idx = np.repeat(np.arange(-512, 512), 1024).reshape(1024,1024).transpose()
+    slot_worst_pix = {}
+    for slot in catalog:
+        corner1 = yagzag_to_pixels(slot['yang'] - slot['halfw'], slot['zang'] + slot['halfw'])
+        corner2 = yagzag_to_pixels(slot['yang'] + slot['halfw'], slot['zang'] - slot['halfw'])
+        row_ranges = (int(corner2[0]) - 1, int(np.ceil(corner1[0])) + 1)
```
jeanconn: This could use a comment indicating I'm explicitly padding by a pixel after rounding up or down.

taldcroft:
Are you going to regenerate the whole database?   What's the plan for using this info?  There is a question of whether it can be viewed as a one-off analysis or for ongoing acq stats modeling.

jeanconn:
I had actually regenerated the whole database over the weekend. (and am doing it again because I had forgotten to record the information about which dark cal the data came from. Granted, it would likely be faster to do this on the side, but the code was more trivial to just toss in here and I think there are fewer things that can go wrong with mapping slots).  And yes, I figured that the outcome of the analysis would help us decide if this has value to be continuously tracked in this kind of way or if it is just a one-off.   I also have the problem that our "bad pixel" list is our list of persistent-over-last-three-dark-currents bad pixel list and the data I've just added is just the warmest pixel from the most recent dark cal.  I may have to check both (persistent bad pixels, recent bad pixels) to see if either has any real value.

taldcroft:
OK on the concept then.  What sort of validation have you done on the indexing to make sure it is right?  There are some famous cases in the ACA status flag analysis of hot pixels being bright enough to be tracked.

jeanconn:
> OK on the concept then. What sort of validation have you done on the
indexing to make sure it is right?

Thus far,  I've just spot checked that 'detected' hot pixels from these
data match indices of known bad pixels in the bad pixel list.

taldcroft:
If we find a driver for this kind of analysis then starting with a full standalone analysis will be the way.  I am not sure that the single brightest pixel is going to strongly inform acq probabilities in a way that is more nuanced than what is already in proseco.  Closing.

PR #90, opened 2016-12-23
Add get_l0_images function for ACA L0 using ACAImage class

taldcroft:
Provide a complementary way to `get_slot_data()` of getting ACA L0 image data.

To do:

- [x] Narrative docs
- [x] Tests
- [x] Remove `main()` stub

### DIFF
--- a/mica/archive/aca_dark/dark_cal.py
+++ b/mica/archive/aca_dark/dark_cal.py
@@ -3,6 +3,7 @@ import os
 from collections import OrderedDict
 import json
 
+import six
 from six.moves import zip
 
 import numpy as np
@@ -10,6 +11,7 @@ from astropy.io import fits
 from astropy.table import Table, Column
 import pyyaks.context
 from Chandra.Time import DateTime
+from chandra_aca.aca_image import ACAImage
 
 from mica.cache import lru_cache
 from mica.common import MICA_ARCHIVE_PATH, MissingDataError
@@ -142,7 +144,50 @@ def get_dark_cal_id(date, select='before'):
 
 
 @DARK_CAL.cache
-def get_dark_cal_image(date, select='before', t_ccd_ref=None):
+def _get_dark_cal_image_props(date, select='before', t_ccd_ref=None, aca_image=False):
+    """
+    Return the dark calibration image (e-/s) nearest to ``date`` and the corresponding
+    dark_props file.
+
+    :param date: date in any DateTime format
+    :param select: method to select dark cal (before|nearest|after)
+    :param t_ccd_ref: rescale dark map to temperature (degC, default=no scaling)
+    :param aca_image: return an AcaImage instance
+
+    :returns: 1024 x 1024 ndarray with dark cal image in e-/s, props dict
+    """
+    DARK_CAL['id'] = get_dark_cal_id(date, select)
+
+    hdus = fits.open(MICA_FILES['dark_image.fits'].abs)
+    dark = hdus[0].data
+    hdus.close()
+
+    with open(MICA_FILES['dark_props.json'].abs, 'r') as fh:
+        props = json.load(fh)
+
+    # Change unicode to ascii at top level
+    if six.PY2:
+        keys = list(props.keys())
+        asciiprops = {}
+        for key in keys:
+            asciiprops[str(key)] = props[key]
+        props = asciiprops
+
+    if t_ccd_ref is not None:
+        # Scale factor to adjust data to an effective temperature of t_ccd_ref.
+        # For t_ccds warmer than t_ccd_ref this scale factor is < 1, i.e. the
+        # observed dark current is made smaller to match what it would be at the
+        # lower reference temperature.
+        t_ccd = props['ccd_temp']
+        dark *= dark_temp_scale(t_ccd, t_ccd_ref)
+
+    if aca_image:
+        dark = ACAImage(dark, row0=-512, col0=-512)
+
+    return dark, props
+
+
+def get_dark_cal_image(date, select='before', t_ccd_ref=None, aca_image=False):
     """
     Return the dark calibration image (e-/s) nearest to ``date``.
 
@@ -152,30 +197,17 @@ def get_dark_cal_image(date, select='before', t_ccd_ref=None):
     :param date: date in any DateTime format
     :param select: method to select dark cal (before|nearest|after)
     :param t_ccd_ref: rescale dark map to temperature (degC, default=no scaling)
+    :param aca_image: return an ACAImage instance instead of ndarray
 
     :returns: 1024 x 1024 ndarray with dark cal image in e-/s
     """
-    DARK_CAL['id'] = get_dark_cal_id(date, select)
-
-    hdus = fits.open(MICA_FILES['dark_image.fits'].abs)
-    dark = hdus[0].data
-    hdus.close()
-
-    if t_ccd_ref is not None:
-        with open(MICA_FILES['dark_props.json'].abs, 'r') as fh:
-            props = json.load(fh)
-        # Scale factor to adjust data to an effective temperature of t_ccd_ref.
-        # For t_ccds warmer than t_ccd_ref this scale factor is < 1, i.e. the
-        # observed dark current is made smaller to match what it would be at the
-        # lower reference temperature.
-        t_ccd = props['ccd_temp']
-        dark *= dark_temp_scale(t_ccd, t_ccd_ref)
-
+    dark, props = _get_dark_cal_image_props(date, select=select, t_ccd_ref=t_ccd_ref,
+                                            aca_image=aca_image)
     return dark
 
 
-@DARK_CAL.cache
-def get_dark_cal_props(date, select='before', include_image=False):
+def get_dark_cal_props(date, select='before', include_image=False, t_ccd_ref=None,
+                       aca_image=False):
     """
     Return a dark calibration properties structure for ``date``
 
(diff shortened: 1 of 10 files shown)


jeanconn:
I had a couple of ideas from some open work on this:
  1) Should the new method work by default in unit labeled e-/sec images instead of the raw DN IMGRAW values?
  2) I'd been playing with some methods to give me the right row col for each pixel a bit more reliably.   Do you have any ideas about including something like that in this framework?

mica/archive/aca_l0.py
```diff
@@ -169,6 +170,63 @@ def get_slot_data(start, stop, slot, imgsize=None,
...
+    if not(np.all(ok)):
+        dat = dat[ok]
+
+    # Convert temperatures from K to degC
+    for temp in ('TEMPCCD', 'TEMPHOUS', 'TEMPPRIM', 'TEMPSEC'):
+        if temp in dat.dtype.names:
+            dat[temp] -= 273.15
+
+    # Masked array col access ~100 times slower than ndarray so convert
+    dat = dat.filled(-9999)
```
jeanconn: Could these get filled with nan values instead without the performance penalty?

taldcroft:
>Should the new method work by default in unit labeled e-/sec images instead of the raw DN IMGRAW values?

From some perspectives you are better off working with the integrated counts in a pixel rather than e-/sec.  Since historically ACA L0 is always DN, I think that within this context it is better to keep that.  Also, you have BGDAVG and RMS that are DN.  But we might well consider putting a `unit` into ACAImage and giving that class the ability to transform easily to e-/sec.

> I'd been playing with some methods to give me the right row col for each pixel a bit more reliably. Do you have any ideas about including something like that in this framework?

Can you clarify?  I have not yet verified against telemetry, but when this is accepted it will be 100% reliable that `img[r, c]` maps to ACA pixel `[r + img.row0, c + img.col0]`.

mica/archive/aca_l0.py
taldcroft: `int` types don't support `NaN`.  

I've also been pondering a different approach which retains the original masked structured array and adds a new column for the `ACAImage` object.  Then one might pass a ref to the corresponding row into `ACAImage` to avoid the expense of converting the row into a dict.  This does create a reference cycle but probably OK.  This approach would eliminate the need for a new function.

jeanconn:
> Can you clarify? I have not yet verified against telemetry, but when this is accepted it will be 100% reliable that img[r, c] maps to ACA pixel [r + img.row0, c + img.col0].

Me, I'd been playing with index/number matrices, because if I need to reshape the image data for a vector operation it is handy to reshape the index matrix too and get those labels tagging along, and it also makes it convenient to mask/fetch by the actual row/col values.  But I can see where that strategy is probably too resource intensive (two extra arrays) for no real benefit to other operations.

taldcroft:
`ACAImage` is meant as a complete replacement for the concept of index/number matrices.   You can always index by actual row/col values trivially.  The problem isn't really resources (these data sets are << Gb, everything is fast enough), but code complexity.

Can you give me an example of why you would "reshape image data for vector operations"?  I'm looking for use cases where a list of `ACAImage` objects is insufficient.

jeanconn:
> ACAImage is meant as a complete replacement for the concept of index/number matrices. 

I'm a little confused.  Did we have index/number matrices hanging around to replace?  

> Can you give me an example of why you would "reshape image data for vector operations"? I'm looking for use cases where a list of ACAImage objects is insufficient.

I'll think more upon this.  I suppose most of my reshape use cases were just broadcasting fails (what's the best way to, say, divide the 8x8s by the possibly-different values of INTEG?), and were trivial enough that no order would have been lost.

mica/archive/aca_l0.py
jeanconn: Regarding "int types don't support NaN." do we have the margin to just use floats?

taldcroft:
> I'm a little confused. Did we have index/number matrices hanging around to replace?

You just said you were "playing with index/number matrices".  So whatever code you were playing with could be replaced with a cleaner formalism (at least that's the goal).

jeanconn:
> You just said you were "playing with index/number matrices". So whatever code you were playing with could be replaced with a cleaner formalism (at least that's the goal).

OK.  I'm not quite sure what the cleaner formalism is for this stuff, but I'll go play more with `ACAImage.aca` and see if it gives me everything I expect.

taldcroft:
Nothing is ever easy... there is a bug in the Ska version of np.ma.MaskedArray:
```
In [14]: arr = np.ma.MaskedArray([None, None])  # Object array

In [15]: arr[1] = np.array([1])  # Put a numpy array into the object

In [16]: arr[1]  # Grr, it got cast into a masked array!
Out[16]: 
masked_array(data = [1],
             mask = False,
       fill_value = 999999)
```
This bug is fixed in 1.11.

taldcroft:
For the record here is a diff that puts ACAImage objects into the output record array.  It will roughly work once we have upgraded Ska numpy, but I'm punting on this approach for now.
```
+
         rowcount += len(chunk)
 
     # just include the rows in the requested time range in the returned data
-    oktime = ((all_rows['TIME'] >= DateTime(start).secs)
-               & (all_rows['TIME'] <= DateTime(stop).secs))
-    return all_rows[oktime]
+    oktime = ((all_rows['TIME'] >= DateTime(start).secs) &
+              (all_rows['TIME'] <= DateTime(stop).secs))
+    all_rows = all_rows[oktime]
+
+    if 'ACAIMG' in columns and 'IMGRAW' in columns:
+        # Predeclare relevant columns for efficiency
+        imgsizes = all_rows['IMGSIZE']
+        imgraws = all_rows['IMGRAW'].filled(0)
+        acaimgs = all_rows['ACAIMG']
+        for ii, row in enumerate(all_rows):
+            imgraw = imgraws[ii].reshape(8, 8, order='F')
+            sz = imgsizes[ii]
+            if sz < 8:
+                imgraw = imgraw[:sz, :sz]
+
+            acaimgs[ii] = ACAImage(imgraw, meta=row)
+
+    return all_rows
 ```

File: mica/archive/aca_l0.py (hunk 1 above)
Comment (taldcroft) [inline mica/archive/aca_l0.py]: > Regarding "int types don't support NaN." do we have the margin to just use floats?

We have margin but it is a bad idea.

**Comment by taldcroft** [comment]
Woohoo, tests!  Computed first moment centroids using get_l0_images and ACAImage machinery and find very close agreement to FM centroids from aspect pipeline.  I'm now highly confident in signs.  

(Note the original used the `F` ordering to reshape from your doc example but that is not correct in the general context.  Not sure if that was just for making the animated GIF come out right?)

**Comment by jeanconn** [comment]
What's your preferred method to get back the row and column of a pixel?

**Comment by taldcroft** [comment]
`img[r, c]` has `row = img.row0 + r`.  Were you thinking of some method like `img.row(r)`?

**Comment by jeanconn** [comment]
I was thinking that your numpy fu was probably better than mine and that you didn't need a method.  Me, my "index/numbers matrix" application was always something like

> In [52]: row_idx, col_idx = np.meshgrid( range(-512, 512), range(-512, 512), indexing='ij')
> In [53]: zip(row_idx[dark_map > 10000], col_idx[dark_map > 10000])
> Out[53]: [(281, 241)]
> In [54]: dark_map.aca[281,241]
> Out[54]: 10970.553

**Comment by taldcroft** [comment]
Rebased on py3 branch and fixed a py3 problem.

**Comment by taldcroft** [comment]
WRT to numpy fu, it is definitely worth googling on things you need to do and then collecting  them in a cheat sheet.  For instance `unravel_index` can be handy, though not in this particular instance.  And then really memorize/understand core things like `np.where` and indexing, which would be the way to do your previous task:
```
In [7]: a = np.arange(16).reshape(4,4)
In [8]: rows, cols = np.where(a > 12)
In [9]: rows, cols
Out[9]: (array([3, 3, 3]), array([1, 2, 3]))

In [10]: a[rows, cols]
Out[10]: array([13, 14, 15])
```

**Comment by taldcroft** [comment]
And then for ACA coordinates:
```
rows = dark_map.row0 + r
cols = dark_map.col0 + c
```

**Comment by taldcroft** [comment]
@jeanconn - I think this PR is good to go now.  I propose merging this and then revisiting #92 and https://github.com/sot/chandra_aca/pull/20.  I think these just need a final sanity-check look-through IIRC.  I didn't bump the version here in anticipation of releasing after #92.

File: mica/archive/aca_dark/dark_cal.py [hunk 2]
```diff
@@ -142,40 +144,70 @@ def get_dark_cal_id(date, select='before'):
...
 
 @DARK_CAL.cache
-def get_dark_cal_image(date, select='before', t_ccd_ref=None):
+def _get_dark_cal_image_props(date, select='before', t_ccd_ref=None, aca_image=False):
     """
-    Return the dark calibration image (e-/s) nearest to ``date``.
-
-    If ``select`` is ``'before'`` (default) then use the first calibration which
-    occurs before ``date``.  Other valid options are ``'after'`` and ``'nearest'``.
+    Return the dark calibration image (e-/s) nearest to ``date`` and the corresponding
```
jeanconn: This docstring seems inconsistent with the "select='before'" default.

taldcroft: "Nearest", subject to the constraint of `select`.

jeanconn: One of the `select` constraints *is* `nearest`, so this isn't perfect.  I understand, but if this dunder method has a bunch of stuff copied over from the public method in the docstring I don't know why I doesn't deserve the full treatment.

jeanconn:
Looks good by me.  Thanks for preserving the API even if I was basically the only user.

mica/archive/aca_dark/dark_cal.py
taldcroft: I've given up on perfect.  :smile:

taldcroft:
I use this too!!

PR #91, opened 2016-12-28
Python 3 compatibility

taldcroft:
Passing tests except for test_aca_hdr3.py.

Got through most of the docs examples OK.  These should just be put into regression tests where sensible.

### DIFF
--- a/mica/archive/aca_l0.py
+++ b/mica/archive/aca_l0.py
@@ -10,8 +10,9 @@ import numpy.ma as ma
 import argparse
 import collections
 import tables
-from itertools import izip, count
+from itertools import count
 
+from six.moves import zip
 import Ska.DBI
 import Ska.arc5gl
 from Chandra.Time import DateTime
@@ -395,7 +396,7 @@ class Updater(object):
                     startdate)
         # find the index in the cda archive list that matches
         # the first entry with the "start" date
-        for idate, backcnt in izip(ingested_files[::-1]['ingest_date'],
+        for idate, backcnt in zip(ingested_files[::-1]['ingest_date'],
                                    count(1)):
             if idate < startdate:
                 break
@@ -405,7 +406,7 @@ class Updater(object):
         # for the entries after the start date, see if we have the
         # file or a later version
         with Ska.DBI.DBI(**self.db) as db:
-            for file, idx in izip(ingested_files[-backcnt:], count(0)):
+            for file, idx in zip(ingested_files[-backcnt:], count(0)):
                 filename = file['filename']
                 db_match = db.fetchall(
                     "select * from archfiles where "
@@ -632,7 +633,7 @@ class Updater(object):
                 os.makedirs(archdir)
             if not os.path.exists(archfile):
                 logger.debug('mv %s %s' % (os.path.abspath(f), archfile))
-                os.chmod(f, 0775)
+                os.chmod(f, 0o775)
                 shutil.move(f, archfile)
             if os.path.exists(f):
                 logger.info('Unlinking %s' % os.path.abspath(f))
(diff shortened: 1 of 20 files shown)


mica/archive/aca_l0.py
```diff
@@ -206,11 +207,13 @@ def __init__(self, msids, start, stop):
...
-    aca_db = dict(dbi="sybase", server="sybase", user="aca_read")
-    with Ska.DBI.DBI(**aca_db) as db:
-        obspars = db.fetchall("""select * from obspar where obsid = %d
-                                 order by obi""" % obsid)
-    return obspars[0]['tstart'], obspars[0]['tstop']
+    from kadi import events  # Kadi is a big import so defer
+    dwells = events.dwells.filter(obsid=obsid)
+    n_dwells = len(dwells)
+    tstart = dwells[0].tstart
+    tstop = dwells[n_dwells - 1].tstop
```
jeanconn: Might want to also handle non-"contiguous" dwells with a warning or some such.

jeanconn: I don't recall what obsid_times is doing anyway.  Seems like it should have been reading from the mica obspar table if it really cared about obspars.

taldcroft: I think most such warnings would be false alarms from small discontinuities like monitor window commanding.  In any case functional enhancements are mostly out of scope for this PR.

From my reading `obsid_times` is a one-off internal method to get times for aca_l0.  It's only called from one place.  So I think the sybase => kadi replacement code is fine.

jeanconn: Yes.  (Though I didn't mean real multi-dwell obsids as much as the set that kadi might return for obsid=0)

taldcroft: One needs to find a balance between possible corner cases and progress.  :smile:

mica/starcheck/starcheck_parser.py
```diff
@@ -5,11 +5,12 @@
 import re
 import subprocess
 import json
-from itertools import izip, count
+from itertools import count
 import numpy as np
 import asciitable
```
jeanconn: This missing in /proj/sot/ska3/test

taldcroft:
Are these warnings new?  They show up in ska_testr (which runs tests so that warnings show up).  Not a blocker, but I'm wondering if this will generate alerts in ska_watch.
```
mica/vv/tests/test_vv.py::test_run_vv Generating V&V for obsid 2121
Processing aspect interval pcadf090549491N003
WARNING: hdu= was not specified but multiple tables are present, reading in first available table (hdu=1) [astropy.io.fits.connect]
WARNING: UnitsWarning: 'marcsec' did not parse as fits unit: At col 0, Unit 'marcsec' not supported by the FITS standard. Did you mean arcsec? [astropy.units.core]
WARNING: UnitsWarning: 'marcsec/yr' did not parse as fits unit: At col 0, Unit 'marcsec' not supported by the FITS standard. Did you mean arcsec? [astropy.units.core]
```

jeanconn:
Yes, those are from the change from Ska.Table to astropy.table.  I hadn't decided yet about muting them in mica.vv.core.

taldcroft:
We can deal with it later...

taldcroft:
BTW the `hdu` one is trivial, just include it in the call.  I'm not sure if there is an easy way to disable the unit warnings.

taldcroft:
See https://github.com/astropy/astropy/pull/5675, which if accepted will fix the warnings problem.

jeanconn:
@taldcroft Are you planning to merge and tag this PR?  I assume there are others hanging around but I wasn't clear on your process plan.

taldcroft:
My plan is to merge all py3 branches now along with py2-3 for skare.  I might try to find a way to scriptify the tagging.

PR #92, opened 2017-01-04
Move dark cal model and scaling to chandra_aca

jeanconn:
Move dark cal model and scaling to chandra_aca


jeanconn:
This is just some ideas for https://github.com/sot/ska-projects/issues/40

mica/archive/aca_dark/__init__.py
```diff
@@ -1,2 +1,2 @@
 from dark_cal import *
-from dark_model import *
+from chandra_aca.dark_model import *
```
jeanconn: Should this import be replaced with some code that will issue a deprecation warning if the user comes looking for dark_model in mica?

taldcroft:
In a perfect world, yes.  But this would be a bit of a pain so I think we can just let it slide here.

mica/archive/aca_dark/__init__.py
taldcroft: In a perfect world, yes. But this would be a bit of a pain so I think we can just let it slide here.

mica/archive/aca_dark/dark_cal.py
```diff
@@ -9,12 +9,12 @@
 import pyyaks.context
 from Chandra.Time import DateTime
 
+from chandra_aca.dark_model import dark_temp_scale
```
taldcroft: This should include `DARK_SCALE_4C`, the anneal pathfinder notebook imports from here.  Yes, useless and obscure, but there you go.

jeanconn: Good catch.  I hadn't thought to look in that project for dependencies.

taldcroft: Searching in the `sot` org or recursively grepping in  our dependencies/repos area is a decent start for finding dependencies.

jeanconn:
> Searching in the sot org or recursively grepping in our
> dependencies/repos area is a decent start for finding dependencies.
>

​Right.  I don't think anything important on the Python side is hiding in
/proj/sot/ska/analysis without a corresponding repo .​

mica/archive/aca_dark/__init__.py
```diff
@@ -1,2 +1,2 @@
 from .dark_cal import *
-from .dark_model import *
+from chandra_aca.dark_model import *
```
jeanconn: Just realized I need, say, "from mica.archive.aca_dark.dark_model import get_warm_fracs" to work, and it doesn't work with this import.  Also doesn't work if I add a "from chandra_aca import dark_model".

jeanconn: I don't see a way to get "from mica.archive.aca_dark.dark_model import get_warm_fracs" to work from the __init__.py, so I just added back a light dark_model.py .

jeanconn:
@taldcroft I probably need a re-review from you to confirm that adding back a file called `dark_model.py` is acceptable (or that there isn't a smarter way to do this from the `__init__.py`.)

taldcroft:
Good to go now.

jeanconn:
Though now I'm more inclined to update mica, kadi, and chandra_aca on a Monday than a Friday...

PR #94, opened 2017-01-06
More mica.vv fixes for Py3

jeanconn:
More mica.vv fixes for Py3

### DIFF
--- a/mica/vv/core.py
+++ b/mica/vv/core.py
@@ -52,7 +52,7 @@ def parse_obspar(file):
                'r': float,
                's': str}
     try:
-        lines = gzip.open(file).readlines()
+        lines = gzip.open(file, 'rt').readlines()
     except IOError:
         lines = open(file).readlines()
     obs_read = csv.DictReader(lines,
--- a/mica/vv/core.py
+++ b/mica/vv/core.py
@@ -22,7 +22,7 @@ from scipy.stats import scoreatpercentile
 
 import Ska.Numpy
 from Chandra.Time import DateTime
-from Ska.Table import read_table
+from astropy.table import Table
 from Ska.astro import sph_dist
 from Ska.engarchive import fetch
 import six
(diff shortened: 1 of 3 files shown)


jeanconn:
This doesn't yet fix the test relative import problem, but at least more of the mica.vv. functions work.

jeanconn:
I don't presently have a Py3 build area to install into to do great "installed module" tests, so I don't know if removing those relative imports fixes the test problem you noticed or not.

taldcroft:
> I don't presently have a Py3 build area to install into to do great "installed module" tests, so I don't know if removing those relative imports fixes the test problem you noticed or not.

Not necessary!!  In the current setup, local testing is equivalent and so building one-off dev ska's for install-testing a single package is not really needed.  (Still useful for integration testing of course).  I noticed that issue with `python setup.py test`.

jeanconn:
OK.  I had seen your `python setup.py test` but hadn't stored that information.  My `python setup.py test` seems to work OK now (after the other fixes) with the more absolute imports or the relative ones in the test module.  Which do you prefer?

taldcroft:
You were just missing `__init__.py` in the `tests` directory so it didn't know it was a package.

Relative imports are preferred within bundled tests, that way you are guaranteed to be getting the right modules.  This is mostly just important in namespace packages where there is always the possibility of problems, but best to just make this standard practice always.

Fixed in c252b42, passing tests for me in ska flight, ska 3 and ska 2.18.

jeanconn:
Well, not quite *missing* `__init__py` just not being good about checking for uncommitted files! :-)  That was another reason I couldn't reproduce your error.

Thanks!

taldcroft:
Right, the dreaded uncommited file (or mistake in tarball creation setup) is one real use case for install-testing.  So in that case you can always do `python setup.py build` and change dir into the right spot in the build directory and do an `import package; package.test()`.

PR #95, opened 2017-02-01
Change two remaining attribute refs (the rest went in py3 vv updates)

jeanconn:
Change two remaining attribute refs (the rest went in py3 vv updates)


jeanconn:
This fixes an introduced bug when trying to check multiple aspect intervals.

PR #97, opened 2017-02-09
Bugfix: Fix another py3 introduced bug and add tests

jeanconn:
Fix another py3 introduced bug and add tests

### DIFF
--- a/mica/vv/tests/test_vv.py
+++ b/mica/vv/tests/test_vv.py
@@ -30,3 +30,9 @@ def test_run_vv():
     obi = process.get_arch_vv(2121)
     assert np.allclose(obi.info()['sim']['max_d_dy'], 0.002197265625)
 
+def test_run_vv_omitted_slot():
+    obi = process.get_arch_vv(19991, version='last')
+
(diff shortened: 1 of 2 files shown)


mica/vv/tests/test_vv.py
```diff
@@ -30,3 +30,9 @@ def test_run_vv():
     obi = process.get_arch_vv(2121)
     assert np.allclose(obi.info()['sim']['max_d_dy'], 0.002197265625)
 
+def test_run_vv_omitted_slot():
+    obi = process.get_arch_vv(19991, version='last')
```
jeanconn: These don't explicitly check anything, but I don't think that is a problem if really I'm just testing the routine didn't thrown an unhandled exception.

taldcroft: That's fine, but put in a code comment to that effect.  Also, if a value like `obi` is not actually used then just skip it.  This makes pyflakes happier.
```
def test_run_vv_omitted_slot():
    process.get_arch_vv(19991, version='last')
```

PR #98, opened 2017-02-10
Improve catalog queries and fetches

jeanconn:
Improve reliability of catalog fetches by obsid.  Add methods for catalog fetches by date and a convenience to get information about monitor windows.

### DIFF
--- a/mica/starcheck/starcheck.py
+++ b/mica/starcheck/starcheck.py
@@ -363,9 +363,10 @@ def update(config=None):
         db_init_cmds = open(db_sql).read()
         db = Ska.DBI.DBI(dbi='sqlite', server=config['starcheck_db']['server'])
         db.execute(db_init_cmds)
-        # make a touch file with a time before starchecks
-        Ska.Shell.bash("touch -t 199801010000 %s" % (FILES['touch_file']),
-                       env={'MAILCHECK': -1})
+        if not os.path.exists(FILES['touch_file']):
+            # make a touch file with a time before starchecks
+            Ska.Shell.bash("touch -t 199801010000 %s" % (FILES['touch_file']),
+                           env={'MAILCHECK': -1})
     else:
         db = Ska.DBI.DBI(dbi='sqlite', server=config['starcheck_db']['server'])
 
--- a/mica/starcheck/starcheck.py
+++ b/mica/starcheck/starcheck.py
@@ -317,7 +317,9 @@ def get_starcheck_catalog(obsid, mp_dir=None,
     sc_id = db.fetchone("select id from starcheck_id "
                          "where dir = '%s'" % mp_dir)['id']
     sc = {'mp_dir': mp_dir}
-    for d in ['manvr', 'catalog', 'obs', 'warnings']:
+    sc['obs'] = db.fetchone("select * from starcheck_obs where obsid = {} and sc_id = {}".format(
+            obsid, sc_id))
+    for d in ['manvr', 'catalog', 'warnings']:
         sc[d] = db.fetchall(
             "select * from starcheck_%s where obsid = %d and sc_id = %d"
             % (d, obsid, sc_id))
--- a/mica/starcheck/starcheck.py
+++ b/mica/starcheck/starcheck.py
@@ -57,10 +57,10 @@ def get_flickpix_mons(start=None, config=None):
     # Check which ones actually ran or are likely to run
     for mon in mons:
         try:
-            actual_catalog = get_starcheck_catalog_at_date(mon['mp_starcat_time'])
-            if (actual_catalog is not None and actual_catalog['obs']['obsid'] == mon['obsid']
-                    and actual_catalog['mp_dir'] == mon['dir']):
-                statuses.append(actual_catalog['status'])
+            catalog = get_starcheck_catalog_at_date(mon['mp_starcat_time'])
+            if (catalog is not None and catalog['obs']['obsid'] == mon['obsid']
+                    and catalog['mp_dir'] == mon['dir']):
+                statuses.append(catalog['status'])
             else:
                 statuses.append('none')
         except LookupError:
--- a/mica/starcheck/starcheck.py
+++ b/mica/starcheck/starcheck.py
@@ -116,12 +116,18 @@ def get_starcheck_catalog_at_date(date, config=None, timelines_db=None):
     # Use kadi to get the first dwell that *ends* after the given time
     dwells = events.dwells.filter(stop__gte=date, subset=slice(None, 1))
     dwell = dwells[0]
-    # Check the timelines that cover the range between the last dwell and the dwell
-    # for this catalog, that should cover NMM when the catalog is commanded
+    # We want to search for legitimate commanding that would cover the time when a star
+    # catalog would have been commanded for this dwell.  This is generally the time range
+    # between the end of the previous dwell and the beginning of this dwell.  However, if
+    # there are multiple dwells from one maneuver, use the beginning of NMM from that one
+    # maneuver else, use the end of the last dwell.  Don't use nman_start time by default
+    # because that doesn't appear to work if the catalog was commanded in a nonstandard
+    # nmm sequence like dark cal.
+    start_cat_search = dwell.manvr.nman_start if dwell.manvr.n_dwell > 1 else dwell.get_previous().stop
     timelines = timelines_db.fetchall(
-        """select * from timeline_loads where scs < 131
+            """select * from timeline_loads where scs < 131
            and datestop > '{}' and datestart < '{}' order by datestart""".format(
-           dwell.get_previous().stop, dwell.start))
+            start_cat_search, dwell.start))
     for timeline in timelines[::-1]:
         starchecks = db.fetchall(
             """select * from starcheck_obs, starcheck_id
--- a/mica/starcheck/starcheck.py
+++ b/mica/starcheck/starcheck.py
@@ -14,6 +14,7 @@ import Ska.DBI
 from kadi import events
 from astropy.table import Table
 
+from mica.utils import get_timeline_at_date
 from mica.starcheck.starcheck_parser import read_starcheck
 from mica.common import MICA_ARCHIVE
 
@@ -29,16 +30,6 @@ DEFAULT_CONFIG = dict(starcheck_db=dict(dbi='sqlite',
 FILES = dict(data_root=os.path.join(MICA_ARCHIVE, 'starcheck'),
              touch_file=os.path.join(MICA_ARCHIVE, 'starcheck', "starcheck_parser.touch"),
              sql_def='starcheck.sql')
-
-# this doesn't belong in starcheck
-def get_timeline_at_date(date, timelines_db=None):
-    date = DateTime(date).date
-    if timelines_db is None:
-        timelines_db = Ska.DBI.DBI(**DEFAULT_CONFIG['timelines_db'])
-    return timelines_db.fetchone(
-        "select * from timeline_loads where datestop >= '%s' "
-        " and datestart <= '%s' and scs <= 130 order by datestart desc"
-        % (date, date))
 
 
 def get_monitor_windows(start=None, stop=None, min_obsid=40000, config=None):
--- a/mica/starcheck/starcheck.py
+++ b/mica/starcheck/starcheck.py
@@ -209,30 +209,6 @@ def get_mp_dir(obsid, config=None, timelines_db=None):
         raise ValueError("get_mp_dir should find something or nothing")
 
 
-def obsid(obsid, mp_dir=None, config=None):
-    """
-    This appears to be a convenience function used in mica.report.report
-    """
-    if mp_dir is None:
-        mp_dir, status, mp_date = get_mp_dir(obsid)
-    if mp_dir is None:
-        raise LookupError("No starcheck catalog found for {}".format(obsid))
-    if config is None:
-        config = DEFAULT_CONFIG
-    db = Ska.DBI.DBI(dbi='sqlite', server=config['server'])
-    sc = db.fetchone(
-        "select id from starcheck_id where dir = '%s'" % mp_dir)
-    if sc is None:
-        raise LookupError("Catalog for {} missing.  Should already be in mica.starcheck db".format(obsid))
-    sc_id = sc['id']
-    sc = {}
-    for d in ['manvr', 'catalog', 'obs', 'warnings']:
-        sc[d] = db.fetchall(
-            "select * from starcheck_%s where obsid = %d and sc_id = %d"
-            % (d, obsid, sc_id))
-    return sc
-
-
 def ingest_obs(obs, obs_idx, sc_id, st, db, existing=None):
     if existing is not None:
         existing_obs = [eobs for eobs in existing
--- a/mica/starcheck/starcheck.py
+++ b/mica/starcheck/starcheck.py
@@ -179,8 +179,8 @@ def get_starcheck_catalog_at_date(date, starcheck_db=None, timelines_db=None):
 def get_mp_dir(obsid, starcheck_db=None, timelines_db=None):
     """
     Get the mission planning directory for an obsid and some status information.  If the obsid catalog was
-    used more than once (multi-obi or rescheduled after being maneuvered-to but not science observed during
-    an SCS107 vehicle interval), return the directory and details of the last one used on the spacecraft.
+    used more than once (multi-obi or rescheduled after being used in a vehicle-only interval), return the
+    directory and details of the last one used on the spacecraft.
 
     :param obsid: obsid
     :param starcheck_db: optional handle to already-open starcheck database
--- a/mica/starcheck/starcheck.py
+++ b/mica/starcheck/starcheck.py
@@ -56,7 +56,6 @@ def get_flickpix_mons(start=None, config=None):
     statuses = []
     # Check which ones actually ran or are likely to run
     for mon in mons:
-        print mon
         try:
             actual_catalog = get_starcheck_catalog_at_date(mon['mp_starcat_time'])
             if (actual_catalog is not None and actual_catalog['obs']['obsid'] == mon['obsid']
--- a/mica/starcheck/starcheck.py
+++ b/mica/starcheck/starcheck.py
@@ -330,9 +330,9 @@ def get_starcheck_catalog(obsid, mp_dir=None, starcheck_db=None, timelines_db=No
     sc['obs'] = db.fetchone("select * from starcheck_obs where obsid = {} and sc_id = {}".format(
             obsid, sc_id))
     for d in ['manvr', 'catalog', 'warnings']:
-        sc[d] = db.fetchall(
+        sc[d] = Table(db.fetchall(
             "select * from starcheck_%s where obsid = %d and sc_id = %d"
-            % (d, obsid, sc_id))
+            % (d, obsid, sc_id)))
     sc['cat'] = sc['catalog']
     del sc['catalog']
     pred_temp = db.fetchone("select pred_ccd_temp from starcheck_pred_temp "
--- a/mica/starcheck/starcheck.py
+++ b/mica/starcheck/starcheck.py
@@ -88,14 +88,19 @@ def get_starcheck_catalog_at_date(date, starcheck_db=None, timelines_db=None):
     A catalog is defined as applying, in this function, to any time from the end of the
     previous dwell through the end of the dwell in which the catalog was used.
 
+    The returned dictionary has starcheck content including catalog ('cat') and maneuver.
+    The 'status' key has possible values 'ran_pretimelines' if the catalog ran before timelines,
+    'planned' if the catalog is in a not-approved future schedule, 'approved' if the
+    catalog is in an approved future schedule (ingested in timelines/cmd_states),
+    and 'ran' if the catalog was executed.
+
     :param date: Chandra.Time compatible date
     :param starcheck_db: optional handle to already-open starcheck database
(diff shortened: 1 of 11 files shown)


mica/starcheck/starcheck.py
```diff
@@ -37,6 +39,74 @@ def get_timeline_at_date(date, timelines_db=None):
         % (date, date))
 
 
+def get_catalog_at_date(date, config=None, timelines_db=None):
```
jeanconn: I'd really prefer to have this out of something like cmd_states or dynamic cmd_states, but ...

mica/starcheck/starcheck.py
```diff
@@ -26,52 +28,167 @@
...
+        config = DEFAULT_CONFIG
+    date = DateTime(date).date
+    if timelines_db is None:
+        timelines_db = Ska.DBI.DBI(dbi='sqlite',
+                                   server='/proj/sot/ska/data/cmd_states/cmd_states.db3')
+    last_tl = timelines_db.fetchone(
+        "select max(datestop) as datestop from timelines")['datestop']
+    first_tl = timelines_db.fetchone(
+        "select min(datestart) as datestart from timelines")['datestart']
+    db = Ska.DBI.DBI(dbi='sqlite', server=config['server'])
```
jeanconn: All through these changes, I'm not sure how much I really need to use the database context manager.

mica/starcheck/starcheck.py
```diff
@@ -26,52 +28,167 @@
...
 
 # this doesn't belong in starcheck
-def get_timeline_at_date(date):
-    aca_db = Ska.DBI.DBI(dbi='sybase', server='sybase', user='aca_read')
-    return aca_db.fetchone(
+def get_timeline_at_date(date, timelines_db=None):
+    date = DateTime(date).date
+    if timelines_db is None:
+        timelines_db = Ska.DBI.DBI(dbi='sqlite',
+                         server='/proj/sot/ska/data/cmd_states/cmd_states.db3')
```
jeanconn: Good point.

mica/starcheck/starcheck.py
```diff
@@ -26,52 +28,167 @@
...
-        sc = db.fetchall(
+    db = Ska.DBI.DBI(dbi='sqlite', server=config['server'])
+    # I'm ignoring the weird case of SIR obsids or whatever where the following ER keeps
+    # the MON
+    mons = db.fetchall("""select obsid, mp_starcat_time, type, sz, yang, zang, dir
+                          from starcheck_catalog, starcheck_id
+                          where type = 'MON' and obsid > 40000
+                          and starcheck_id.id = starcheck_catalog.sc_id""")
+    mons = Table(mons)
+    if start is not None:
```
jeanconn: OK, I was too lazy to thing through the conditional SQL, but your `start_date` is fine.  And not sure about your 'mp_starcat_time' comment.  It is actually a date, just has a that `time` because it has been around a while (before I was good about date vs time).  I could alias it, rename it in the database or whatever.

Unrelatedly, while I love the comments, I thought "WIP" was the thing to do if I wasn't ready for them.

mica/starcheck/starcheck.py
```diff
@@ -26,52 +28,167 @@
...
+                statuses.append('none')
+        except LookupError:
+            statuses.append('none')
+    # Add that status info to the table and filter on it
+    mons['status'] = statuses
+    mons = mons[mons['status'] != 'none']
+    return mons
+
+
+def get_catalog_at_date(date, config=None, timelines_db=None):
```
jeanconn: Regarding `timelines_db` I was actually thinking about this both because I wanted to be able to run against Sybase or sqlite and because it gave options in terms of optimizing db handle opens (not open closing the file for each query if needed).

mica/starcheck/tests/test_catalog_fetches.py
```diff
@@ -0,0 +1,86 @@
...
+            # This doesn't detect MON just yet
+            cat[slot] = {'type': 'STAR',
+                         'yag': np.median(yags) * 3600.,
+                         'zag': np.median(zags) * 3600.,}
+    return cat, telem
+
+
+def test_validate_catalogs_over_range():
+    start = '2017:001'
+    stop = '2017:005'
```
jeanconn: This test could be parametrized to check a few smaller intervals.  Right now, dwells during safing action or recovery will fail.  Most observations during a vehicle-only interval should be OK.

mica/starcheck/starcheck.py
```diff
@@ -20,60 +22,197 @@
...
+    """
+    For a given date, return a dictionary describing the starcheck catalog that should apply.
+    The content of that dictionary is from the database tables that parsed the starcheck report.
+    A catalog is defined as applying, in this function, to any time from the end of the
+    previous dwell through the end of the dwell in which the catalog was used.
+
+    :param date: Chandra.Time compatible date
+    :param config: optional mica config which can override starcheck sqlite database, mp dir, or timelines db
+    :param timelines_db: optional handle to already-open timelines database
+    :returns: dictionary with starcheck content including catalog and maneuver
```
jeanconn: Should probably define a convention for the 'status' values being used.

mica/starcheck/tests/test_catalog_fetches.py
```diff
@@ -0,0 +1,86 @@
...
+    cmd_q4 = np.sqrt(np.abs(1 - cmd_quats['AOCMDQT1'].vals[0]**2
+                            - cmd_quats['AOCMDQT2'].vals[0]**2
+                            - cmd_quats['AOCMDQT3'].vals[0]**2))
+    return Quat(normalize([cmd_quats['AOCMDQT1'].vals[0],
+                                      cmd_quats['AOCMDQT2'].vals[0],
+                                      cmd_quats['AOCMDQT3'].vals[0],
+                                      cmd_q4]))
+
+
+def get_trak_cat_from_telem(date, cmd_quat):
```
taldcroft: Shouldn't this be in `chandra_aca`?

taldcroft: `trak_cat` is not really standard terminology.  `starcat`?

taldcroft: Requiring a separate call to `get_cmd_quat` seems inconvenient in the API.  Is there a reason for breaking this out?  You can at least make the parameter be `cmd_quat=None` and then get it if None.  But agreed it might be useful to maintain `get_cmd_quat` as a separate utility function in chandra_aca.

mica/starcheck/tests/test_catalog_fetches.py
```diff
@@ -0,0 +1,86 @@
...
+                         'zag': np.median(zags) * 3600.,}
+    return cat, telem
+
+
+def test_validate_catalogs_over_range():
+    start = '2017:001'
+    stop = '2017:005'
+    dwells = events.dwells.filter(start, stop)
+    for dwell in dwells:
+        telem_quat = get_cmd_quat(dwell.start)
```
taldcroft: Naming nit.  This should be `cmd_quat` as opposed to the ambiguous `telem_cat`.

mica/starcheck/tests/test_catalog_fetches.py
```diff
@@ -0,0 +1,86 @@
...
+                                       telem['AOACZAN{}'.format(slot)].vals[track][sample] / 3600.,
+                                       qref)
+                yag, zag = radec2yagzag(ra, dec, cmd_quat)
+                yags.append(yag)
+                zags.append(zag)
+            # This doesn't detect MON just yet
+            cat[slot] = {'type': 'STAR',
+                         'yag': np.median(yags) * 3600.,
+                         'zag': np.median(zags) * 3600.,}
+    return cat, telem
```
taldcroft: Is there a use-case for needing all the telemetry as well?

mica/starcheck/tests/test_catalog_fetches.py
jeanconn: `starcat` is ambiguous for me as I'm not sure if it is acquisition or kalman or both.  The "tracked" things is what I was shooting for here.

jeanconn: How much do we care about the testing API?  That was my thought at the time.

mica/starcheck/tests/test_catalog_fetches.py
jeanconn: Every time the answer didn't come out as expected, I wanted full telemetry to see why without using the debugger.  For testing, this was a win.

mica/starcheck/tests/test_catalog_fetches.py
jeanconn: I was shooting for the telemetered commaned quat as opposed to the commanded quaternion from cmd states.

mica/starcheck/starcheck.py
```diff
@@ -20,60 +22,197 @@
...
+                                        server=os.path.join(os.environ['SKA'], 'data', 'cmd_states', 'cmd_states.db3')))
 FILES = dict(data_root=os.path.join(MICA_ARCHIVE, 'starcheck'),
              touch_file=os.path.join(MICA_ARCHIVE, 'starcheck', "starcheck_parser.touch"),
              sql_def='starcheck.sql')
 
 # this doesn't belong in starcheck
-def get_timeline_at_date(date):
-    aca_db = Ska.DBI.DBI(dbi='sybase', server='sybase', user='aca_read')
-    return aca_db.fetchone(
+def get_timeline_at_date(date, timelines_db=None):
```
taldcroft: Maybe need a new `mp_utils` module in `mica` to collect up random stuff like this.  Or just `utils`?

mica/starcheck/tests/test_catalog_fetches.py
jeanconn: Regarding "shoudn't this be in `chandra_aca`, again I was thinking of this as a local method in testing with exceptions defined just for this testing etc.   I was not looking to spec the more general method.   More general method on my todo list for `chandra_aca`, but the exceptions were picking at me.

mica/starcheck/starcheck.py
jeanconn: mica.utils or should I had some helper methods to either Chandra.cmd_states or a new timelines module?

mica/starcheck/starcheck.py
```diff
@@ -20,60 +22,197 @@
...
+    if timelines_db is None:
+        timelines_db = Ska.DBI.DBI(**DEFAULT_CONFIG['timelines_db'])
+    return timelines_db.fetchone(
         "select * from timeline_loads where datestop >= '%s' "
         " and datestart <= '%s' and scs <= 130 order by datestart desc"
         % (date, date))
 
-def get_mp_dir(obsid, config=None):
+
+def get_flickpix_mons(start=None, config=None):
```
jeanconn: Also, I was originally going to move this out of here because it is very specific to the task at hand, but on the other... hand, it *is* really just a special star catalog lookup, so maybe it isn't dreadful to just leave it in the star catalog lookup utilities.

taldcroft: OK.  @malgosias and I are discussing this in the context of scheduling flickering pixels.  We need to be able to run this function on GRETA, which means keeping a version of the `starcheck.db3` database on GRETA.  Would that be reasonable, and are there any other dependencies?  (Now I forget if cmd_states.db3 is actually being synced as well).

jeanconn:
It might be easier in the short term to run this function on HEAD to make
short table of the run or about-to -be-run monitor windows, and then sync
just the table.

We've been shooting to get copies of all the good stuff over to GRETA
though, so I'm also onboard with that.

mica/starcheck/starcheck.py
jeanconn: (sticking this in as a code comment instead of top-level PR comment) It might be easier in the short term to run this function on HEAD to make a short table of the run or about-to -be-run monitor windows, and then sync just the table.

We've been shooting to get copies of all the good stuff over to GRETA
though, so I'm also onboard with that.

mica/starcheck/starcheck.py
```diff
@@ -20,60 +22,197 @@
...
+                          where type = 'MON' and obsid > 40000
+                          and mp_starcat_date >= '{}'
+                          and starcheck_id.id = starcheck_catalog.sc_id""".format(start_date))
+        mons = Table(mons)
+    with Ska.DBI.DBI(**config['timelines_db']) as timelines_db:
+        statuses = []
+        # Check which ones actually ran or are likely to run
+        for mon in mons:
+            try:
+                catalog = get_starcheck_catalog_at_date(mon['mp_starcat_date'],
```
taldcroft: Just used this function for a real-life investigation and I immediately wanted this `catalog` for each mon.  So you could make an empty `catalog` column with `dtype=object` (using `mons['catalog'] = None`)
 and then set below with `mons[i_mon] = catalog` (and add `i_mon` via enumerate).

mica/starcheck/starcheck.py
```diff
@@ -20,60 +22,197 @@
...
+    # I'm ignoring the weird case of SIR obsids or whatever where the following ER keeps
+    # the MON
+    start_date = DateTime(start or '1999:001').date
+    with Ska.DBI.DBI(dbi='sqlite', server=config['server']) as db:
+        mons = db.fetchall("""select obsid, mp_starcat_time as mp_starcat_date, type, sz, yang, zang, dir
+                          from starcheck_catalog, starcheck_id
+                          where type = 'MON' and obsid > 40000
+                          and mp_starcat_date >= '{}'
+                          and starcheck_id.id = starcheck_catalog.sc_id""".format(start_date))
+        mons = Table(mons)
```
taldcroft: Move this out of the context manager.

mica/starcheck/starcheck.py
taldcroft: Is there a credible code path that hits Sybase (i.e. will this function always work with just `cmd_states.db3` and `starcheck.db3`)?  In general I'd rather just sync the source database tables instead of a special-case derived product.

jeanconn: No, I didn't have that as a concern.  I think you could only end up with a Sybase query if you had mismatched versions of this code.  I just haven't played much with running any of this mica code on GRETA so didn't want to accidentally extend the implementation timeline.

mica/starcheck/tests/test_catalog_fetches.py
taldcroft: OK.

taldcroft: I was thinking about this being a public method somewhere else.  But yes, it can be left here and API doesn't matter then.

mica/starcheck/tests/test_catalog_fetches.py
taldcroft: OK.

mica/starcheck/tests/test_catalog_fetches.py
taldcroft: OK.

mica/starcheck/starcheck.py
taldcroft: Let's go with `mica.utils`.  Not perfect but easy and good enough.

mica/starcheck/starcheck.py
taldcroft: This function will ultimately be a dependency of the flickering pixel project and will be used to update the list of already-run monitor windows.  It doesn't have to be made GRETA-ready for this PR but eventually yes.

mica/starcheck/starcheck.py
taldcroft: It's good to use when possible (ensures nicely closing the DB) but this code is OK.

mica/starcheck/starcheck.py
taldcroft: Just looking at this again it seems like it would be useful and easy to do three things:
- Add a stop time.
- Add a `min_obsid=40000` parameter to allow getting all monitor windows if desired.
- Rename to `get_monitor_windows` as a more general function.

taldcroft:
My review is finally done.  Improved docs (narrative docs) can wait.

jeanconn:
Your initial comments brought up a bunch of things I'd like to clean up here; my plan was to expedite those to clean this up before we start increasing dependence on it (especially differences in the fetches by obsid and date).

mica/starcheck/starcheck.py
```diff
@@ -20,60 +22,197 @@
...
+               and starcheck_id.id = starcheck_obs.sc_id
+               order by mp_starcat_time """.format(
+                timeline['mp_dir'],
+                timeline['datestart'],
+                timeline['datestop'], dwell.start))
+        # The last one should be the one before beginning of the dwell
+        if len(starchecks):
+            cat_info = get_starcheck_catalog(starchecks[-1]['obsid'], mp_dir=starchecks[-1]['dir'])
+            if cat_info['obs']['mp_starcat_time'] > DateTime().date:
+                cat_info['status'] = 'approved'
```
jeanconn: Oops.  If the obsid it is in timelines but not in telemetry/kadi yet, this isn't going to work anyway (because we've already assumed finding a dwell earlier in this code).  I've forgotten, what's the fastest kadi way to get the last available dwell?

mica/starcheck/starcheck.py
jeanconn: A bit confused by the implementation suggestion here.  Did you mean `mons[i_mon]['catalog'] = catalog` or are you saying you want me to replace the entries of mons with catalogs?  And if you meant `mons[i_mon]['catalog'] = catalog` why would I need to use `i_mon` to index to the row instead of just making the assignment in the row as `mon['catalog'] = catalog`?

jeanconn:
I note that I've now hit acq_stats and the ingest processing for the mica starcheck database, so both will need more explicit tests.  Compatibility with chandra_aca.plot.plot_stars should also be explicitly tested.

mica/starcheck/tests/test_catalog_fetches.py
```diff
@@ -0,0 +1,86 @@
...
+                raise ValueError("Unexpected offset in pointing")
+            else:
+                print(dwell.start, "pointing offset but has unknown manvr template")
+                continue
+        trak_sc = sc['cat'][sc['cat']['type'] != 'ACQ']
+        for slot in range(0, 8):
+            if slot in cat:
+                slot_match = trak_sc[trak_sc['slot'] == slot]
+                if not len(slot_match):
+                    raise ValueError("missing slot in catalog")
```
jeanconn: I note that when run over a larger chunk of time, this if statement caught obsids that had fewer than 8 slots commanded during guide.  It looks like data in "AOACFCT=TRAK" status with "AOIMAGE=STAR" set is occasionally seen in a slot that was last used for acq and then unassigned.  Obsids include 54703, 54381, 51599, 51593, 51591, 51589, 51450.

mica/starcheck/tests/test_catalog_fetches.py
```diff
@@ -0,0 +1,86 @@
...
+                trak_sc_slot = slot_match[0]
+                if trak_sc_slot['type'] == 'FID':
+                    assert cat[slot]['type'] == 'FID'
+                assert np.abs(cat[slot]['yag'] - trak_sc_slot['yang']) < 20.0
```
jeanconn: It looks like this code that was doing the position checking needed a 20 arcsec limit because I was sharing the same limit between fids and stars, and the fids have "standard" offsets at this time.  I'll set different limits so this code captures that understanding.

jeanconn:
I think https://github.com/sot/mica/compare/d9458781...9de36a7?expand=1 shows a diff of just the changes since you said this was ok...

mica/starcheck/starcheck.py
```diff
@@ -8,90 +8,234 @@
...
+                          and mp_starcat_date >= '{}'
+                          and mp_starcat_date <= '{}'
+                          and starcheck_id.id = starcheck_catalog.sc_id""".format(
+                min_obsid, start_date, stop_date))
+    mons = Table(mons)
+    mons['catalog'] = None
+    with Ska.DBI.DBI(**config['timelines_db']) as timelines_db:
+        statuses = []
+        # Check which ones actually ran or are likely to run
+        for i_mon, mon in enumerate(mons):
```
jeanconn: Oops, looks like enumerated but didn't wait to commit this before finding out from @taldcroft why that was suggested as part of the strategy to add the catalogs to the data structure.

jeanconn:
For testing `chandra_aca.plot` I confirm that this one test worked fine:
```
sc = mica.starcheck.starcheck.get_starcheck_catalog(19990)
chandra_aca.plot.plot_stars([211.929783, 44.465961, 110.897312], catalog=sc['cat'])
```
And I note that the docs for `chandra_aca.plot.plot_stars` indicate that any astropy.table compable structure for catalog will do.  The new table from get_starcheck_catalog *is* an astropy.table so explicitly compatible.

taldcroft:
Docs on `get_mp_dir` don't say what the return value `date` refers to.  Also the `status` values can be more readably described as a list in the main docstring (above the `:params:` and `:returns:`.

jeanconn:
Oh, you probably meant a pretty list...

taldcroft:
Hmm, that surprises me since it contradicts the RST spec (see e.g. http://docutils.sourceforge.net/docs/user/rst/quickref.html#literal-blocks) that requires indentation.  Scroll up a bit for an example of bulleted list.

jeanconn:
​Maybe that was a sphinx weirdness?  Either way, I just removed the double
"::" and the real bullets look fine as well.​

taldcroft:
I didn't notice before, but it appears (from the docs) that the acquisition probability is not included.  Is that right?  That would be extremely useful, both for convenience and for a historical record of the as-planned probability.

Issue #104, opened 2017-03-21
Mica.starcheck fails in Ska3 (no asciitable)


taldcroft:
This would indicate there is inadequate test coverage and/or tests weren't run on Py3.

jeanconn:
Tests weren't run on Py3.  This code has not had Py3 conversion attention yet, and that wasn't on top of my list for #98.  I had a separate branch open working the Py3 fixes and updating tests.

jeanconn:
Closed by #105

Issue #103, opened 2017-03-20
Add acquisition probability to database and fetches


jeanconn:
Add acquisition probability to database and fetches

jeanconn:
Closed by #105

PR #105, opened 2017-03-22
Fix some more Py3 issues and add P_ACQ to the starcheck database

jeanconn:
Fix some more Py3 issues and add P_ACQ to the starcheck database.  The P_ACQ item should probably be in another pull request, but it just ended up in continuity with the fixes to mica.starcheck.  In there I also pulled out the "ingest/update/process" bits into their own file.


jeanconn:
Closes #104

jeanconn:
I think this also Closes #103

jeanconn:
I was hoping to put this and #106 and #107 and install ... now.

PR #108, opened 2017-03-24
Replace shell "find" with os.walk

jeanconn:
Replace shell "find" with os.walk in starcheck ingest

This code was set to try to find and ingest "new" starcheck.txt files using a Linux shell
"find" command set to find only files newer than the last ingested file. The time of the
last ingested file was stored by updating *another* file with that file's mod or change
time. The "find" command syntax was ugly, doing this in the shell was also not pretty, and
using a separate tracking file added a place where this could go wrong. This commit adds
the file mod time to the database (so an external tracking file isn't needed) and then
uses os.walk to get files and their mod times to include or exclude them from an update.
The os.walk code is borrowed from kadi.events.orbit_funcs .

The new P_ACQ column requires a database rebuild anyway, and it looks like the "find" command
is responding a bit differently than it has in the past (not sure if files in /data/mpcrit1/mplog have had
permissions changes more recent than their last mod time), so this would be good to just get done now.

### DIFF
--- a/mica/starcheck/process.py
+++ b/mica/starcheck/process.py
@@ -107,7 +107,7 @@ def prune_dirs(dirs, regex):
 def get_new_starcheck_files(rootdir, mtime=0):
     """
     Look for starcheck.txt files in a a given top-level SOT MP directory
-    and return those with modification times after the optional supplied 'mtime'.
+    and return those with modification times at or after the optional supplied 'mtime'.
     """
     logger.info("Getting new starcheck.txt files from {}".format(rootdir))
     starchecks = []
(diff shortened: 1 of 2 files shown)


taldcroft:
I actually stared at that `find` command for a while but stopped myself from asking you to rewrite it.  But definitely :+1: on this!

jeanconn:
Nah, you'd asked me to rewrite it as os.walk in 2013 when I first started
this :-)

jeanconn:
My plan was to remake the database and just check that the new one has the
same number of entries.

jeanconn:
From a review of the rebuilt database using this os.walk, there are benign differences:

   * The new table doesn't have entries for JUL1009  .  It looks like JUL1009 at some point was made into a symbolic link to JUL1309.  Not really an issue for our downstream use of any data in that "one" directory.
   * The new table doesn't have an entry for MAY1915, which doesn't seem to exist at this point in the MP area.  I think that might have been a test directory that was labeled incorrectly and then fixed.

I'm good with these changes.

PR #110, opened 2017-03-29
Add a configurable start time for ingest of starcheck.txt files in the database

jeanconn:
Add a configurable start time for ingest of starcheck.txt files

#108 removed the need for the silly "touch file".  That was a file manually set with a timestamp and it was used in combination with an option to the shell "find" command to find files newer that that timestamp matching the pattern.  That was used to get "new" starcheck.txt files to update the database.  It was also how one could define a more recent time to, say, just get some recent files to  make a quick testing database.  This PR just adds a "start" time to the utility to have that option for testing.  Theoretically, I could also have just made a small /data/mpcrit1/ or used some similar technique to get just a handful of files for testing.

### DIFF
--- a/mica/starcheck/process.py
+++ b/mica/starcheck/process.py
@@ -91,7 +91,7 @@ def get_options():
     parser.add_argument("--mp-top-level",
                         help="top level SOT MP dir")
     parser.add_argument("--start",
-                        help="update database with starcheck files after this strt time")
+                        help="update database with starcheck files after this start time")
     opt = parser.parse_args()
     return opt
 
--- a/mica/starcheck/process.py
+++ b/mica/starcheck/process.py
@@ -90,6 +90,8 @@ def get_options():
                         help="server or file for database")
     parser.add_argument("--mp-top-level",
                         help="top level SOT MP dir")
+    parser.add_argument("--start",
+                        help="update database with starcheck files after this strt time")
     opt = parser.parse_args()
     return opt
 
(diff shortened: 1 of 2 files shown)


mica/starcheck/process.py
```diff
@@ -149,9 +151,11 @@ def update(config=None):
     else:
         db = Ska.DBI.DBI(dbi='sqlite', server=config['starcheck_db']['server'])
         max_mtime = db.fetchone("select max(mtime) as mtime from starcheck_id")
-        if max_mtime is not None:
+        if max_mtime is not None and 'start' not in config:
```
taldcroft: Since it isn't computationally expensive, you can revert this change and just have the override at the end.  This simplifies the logic.

mica/starcheck/process.py
```diff
@@ -149,9 +151,11 @@ def update(config=None):
...
             last_starcheck_mtime = max_mtime['mtime']
-
+    # if a start time is explicitly requested, override 0 or last database value
```
taldcroft: Capitalize comments most of the time.

jeanconn:
Never too late.  I just figured when I didn't hear from you for 2 days on the date format thing that I should try to just do something with mica and be done.

mica/starcheck/process.py
jeanconn: Good point.   This is basically a benign bug missed in cleanup.

taldcroft:
When you get a breather (maybe never), you should start trying to de-`config` (and de-`DEFAULT_CONFIG`) mica.  It makes it so hard to read the code and figure out what is happening.  Here is what I just went through.  Sorry for the stream of thought and slight rant.

Start looking at code, and try to check where `config['start']` is coming from.  OK, it's getting options from `get_options` and passing those to `update`.
```
def main():
    """
    Command line interface to update starcheck database from files
    in MP directories
    """
    opt = get_options()
    update(config=vars(opt))
```
But wait, early in update it is looking for `config['starcheck_db']['dbi']`.  Hummm, that means that somehow a dict is coming out of `get_options`?  What??  After some bouncing around I see that it is setting defaults for items that are not actually part of the option parser.  Weird, does that actually work, apparently so.

But now what is happening with the new `start` config option.  Looking at this it seems like `config` should *always* have a `start` since that is one of the options.  But maybe by some chance `vars(opt)` doesn't include `start` if `--start` wasn't explicitly included, even though `opt.start` is guaranteed to exist and has a value of `None`?

Takeaway: if the function options are all explicitly stated with obvious defaults, and `get_options` only gets what it is advertising, then code readers have an easier time.

jeanconn:
> Sorry for the stream of thought and slight rant.

No no, I was thinking the same thing in here when it was becoming more obvious to me that I didn't know which places I needed to set/handle a new `start`, and then trying to set it without using get_options() was just a mess.

PR #118, opened 2017-05-05
Get osol from archive when fetching asp1

jeanconn:
Get osol from CXCDS archive when fetching other asp1 content.

This appears to not be included with the standard "get asp1", so it needs an additional "get".

### DIFF
--- a/mica/archive/asp_l1.py
+++ b/mica/archive/asp_l1.py
@@ -34,7 +34,7 @@ CONFIG = dict(data_root=os.path.join(MICA_ARCHIVE, 'asp1'),
               full='asp1',
               filecheck=False,
               cols=ARCHFILES_HDR_COLS,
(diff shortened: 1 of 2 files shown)


jeanconn:
I think the easiest way to update the archive with all the content is to rebuild the asp1 archive by moving its directory and doing  `update_asp_l1.py --rebuild`.

jeanconn:
I tested this code change explicitly on one observation and confirmed it grabbed the right osol files .  I had forgotten that the osol files are not knit back into one file to cover the aspect interval, though, so tools wanting the osol for an observation will need to read all the files.  Granted, it may be that we don't want this anyway and it is easier to just use AOATTQT over the intervals.

taldcroft:
Closing this per discussions that the `osol` files may have inconsistent / wrong misalign corrections applied.  We can re-open if this changes or if a new driver appears.

PR #100, opened 2017-02-23
First draft of guide stats 

jeanconn:
First draft of guide stats (based on acq stats)

PR #119, opened 2017-05-19
First draft of guide stats

jeanconn:
First draft of guide stats for mica.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -110,7 +110,7 @@ Functions
 :mod:`mica.utils`
 =================
 
-..automodule: mica.utils
+.. automodule:: mica.utils
 
 Functions
 ---------
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -486,10 +486,10 @@ and methods to retrieve content from that database.  The `mica.starcheck` databa
 inherits much of the structure of the original Sybase starcheck_database project which
 organized the per-obsid data into these tables:
 
-   starcheck_obs
-   starcheck_manvr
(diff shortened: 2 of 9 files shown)


jeanconn:
This content is from #100.  See discussion and approval there.

taldcroft:
Sorry, I realized there are a couple more things:
- Validation of results?  This is kind of a big thing, but may at least a few plots showing new vs. old for similar quantities?
- Documentation of the new table in mica docs.

taldcroft:
Also you should make a habit of deleting branches as soon as you merge.  There is a bunch of cruft now in mica.

jeanconn:
Right, a couple of things.  

1) I was refraining from deleting branches because I've just been too tired.   As in, I've been making mistakes and it can be easier to undo them if I leave the branches for more than a couple of minutes after merging.  So I've been going back and deleting branches after a couple of days.  I think the only really stale branch is the vv-comments branch and I don't know what that is.

2) I had validation, fixing Py3 compatibility in my access methods, and adding docs on my list, but wasn't sure if I should just install and start updating the table independently.  I suppose I can just wait.  I had just pushed a mica 3.10 update to skare master but I'll rebase without that and finish up these items before installing then.

taldcroft:
OK on your deleting strategy.  I never remember to do anything 2 days later unless I leave a reminder.

I wasn't intending on the follow-on to stop this from being part of production!  Definitely want to see the table available and being updated.  Maybe open some issues for now.

Issue #121, opened 2017-05-19
Guide stats: Py3


jeanconn:
The `get_stats` methods in acq and guide stats are using tables.openFile and need to be made Py3 compatible.

jeanconn:
Closed by #125

PR #125, opened 2017-05-19
Read and write guide and acq stats with tables3_api

jeanconn:
Read guide and acq stats with Py3-compat tables_open_file.

### DIFF
--- a/mica/stats/acq_stats.py
+++ b/mica/stats/acq_stats.py
@@ -9,6 +9,7 @@ from Chandra.Time import DateTime
 import mica.archive.obspar
 from mica.starcheck import get_starcheck_catalog, get_starcheck_catalog_at_date
 import tables
+import tables3_api
 import numpy as np
 import Ska.quatutil
 import Ska.astro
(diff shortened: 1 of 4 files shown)


jeanconn:
Closes #121 (to at least make *reading* the files Py3 compatible)

taldcroft:
What about making a new module `tables3` that includes code like below in its `__init__.py` to monkey patch `tables`:
```
import tables
from distutils.version import LooseVersion

if LooseVersion(tables.__version__) < LooseVersion('3.0'):
    tables.open_file = tables.openFile
    ...
```
Then any code that needs to work with `tables` 2.x and 3.x can do so with:
```
import tables
import tables3  # inject small subset of tables 3.x API to tables

# Use the 3.x API
```
Eventually we will have migrated to Py3 with tables 3.x and the `import tables3` statements can be dropped.

jeanconn:
Sure.  I thought we'd talked about monkey patching (and just patching) and decided to pass, but if you think this will reduce our maintenance headache, I'm good with it.

jeanconn:
Also, do we just need this for open_file?  What about readWhere or others that I'm not immediately remembering?  readWhere would be a Table method and I'm not sure how deep you want to patch.

taldcroft:
Well at least give it a try with the Table methods you can think of, starting with `readWhere`.

jeanconn:
So far I've got readWhere, getWhereList, and readCoordinates

https://github.com/sot/tables3/pull/1/files

PR #126, opened 2017-05-25
In V&V, handle case of uncommanded image window during Kalman

jeanconn:
In V&V, handle case of uncommanded image window during Kalman.  The previous code assumed that any missing slots had entries in the ocat stars table.  This change checks to see if missing slots have entries in the table.  If a slot has no entry, it is just skipped.  This is designed for catalogs that are commanded with fewer than 8 used image windows during Kalman.


jeanconn:
Skipping request of explicit @taldcroft review for this tiny bugfix.

PR #131, opened 2017-05-30
Add acquisition anomaly warnings

jeanconn:
The current email acquisition anomaly warnings are from the star_stats_db processing (which is the Python code that is updating the Sybase tables).  This PR adds an email warning for acquisition anomalies to mica acq_stats.  Unlike the current star_stat_db code, it actually checks to see if the data shows a "classic" acquisition anomaly or just a star with a large deviation from expected position.  Also unlike the star_stat_db code, it uses the halfwidth to check to see if the deviation is somewhat anomalous.

### DIFF
--- a/mica/stats/acq_stats.py
+++ b/mica/stats/acq_stats.py
@@ -15,6 +15,8 @@ import Ska.astro
 from mica.quaternion import Quat
 from chandra_aca import dark_model
 import logging
+import smtplib
+from email.mime.text import MIMEText
 
 import warnings
 # Ignore known numexpr.necompiler and table.conditions warning
@@ -104,7 +106,7 @@ ACQ_COLS = {
 
 SKA = os.environ['SKA']
 table_file = os.path.join(SKA, 'data', 'acq_stats', 'acq_stats.h5')
-
+DEFAULT_EMAIL = ['aca_alert@cfa.harvard.edu']
 
 def get_options():
     parser = argparse.ArgumentParser(
@@ -114,6 +116,10 @@ def get_options():
                         help="check for missing observations in table and reprocess")
     parser.add_argument("--obsid",
                         help="specific obsid to process.  Not required in regular update mode")
+    parser.add_argument('--email',
+                        action="append",
+                        help="email warning recipient, specify multiple times "
+                        + "for multiple recipients")
     opt = parser.parse_args()
     return opt
 
@@ -441,6 +447,50 @@ def _get_obsids_to_update(check_missing=False):
     return obsids
 
 
+def warn_on_acq_anom(acqs, emails):
+    """
+    Log and email warning about any acquisition stars with observed positions outside the
+    expected search box (plus a pad).  For acquisition stars with tracked positions in the wrong
+    search box, note the box (classic acquisition anomaly).
+
+    :param acqs: astropy table of acquisition stats, includes expected and observed position,
+                 obsid, slot, and halfw used in this method.
+    :param emails: list of addresses to receive email warning if any are generated
+    """
+    # Find tracked objects in the acq stats table outside the intended search box plus padding
+    # img_func == NONE means nothing was tracked
+    box_pad = 16  # arcsecs
+    anom_match = ((acqs['img_func'] != 'NONE') &
+                  ((np.abs(acqs['yang_obs'] - acqs['yang']) >= (acqs['halfw'] + box_pad))
+                   | (np.abs(acqs['zang_obs'] - acqs['zang']) >= (acqs['halfw'] + box_pad))))
+    for anom in acqs[anom_match]:
+        # Check to see if the star is actually found in another box.
+        other_box_match =  ((np.abs(anom['yang_obs'] - acqs['yang']) <= (acqs['halfw'] + box_pad))
+                            & (np.abs(anom['zang_obs'] - acqs['zang']) <= (acqs['halfw'] + box_pad)))
+        if np.any(other_box_match):
+            text = "Acquisition Anomaly.  Star for slot {} actually in box {} \n".format(
+                anom['slot'], acqs[other_box_match][0]['slot'])
+        else:
+            text = "Does not appear to be classic star-in-wrong-box anomaly\n"
+        # Make a dictionary of the anom record for use in string formatting
+        output_dict = {col: anom[col] for col in anom.colnames}
+        output_dict['dy'] = anom['yang_obs'] - anom['yang']
+        output_dict['dz'] = anom['zang_obs'] - anom['zang']
+        text += """Large Deviation from Expected ACQ Star Position in {obsid}
+      Slot {slot} Expected (Y-Pos, Z-Pos) = ({yang:.1f}, {zang:.1f})
+      Slot {slot} Observed (Y-Pos, Z-Pos) = ({yang_obs:.1f}, {zang_obs:.1f})
+      Halfwidth {halfw:03d}        (dy, dz) = ({dy:.1f}, {dz:.1f})""".format(
+            **output_dict)
+        # Log and Send message for slot.  Obsid can have more than one email
+        logger.warn(text)
+        msg = MIMEText(text)
+        msg['From'] = 'aca@head.cfa.harvard.edu'
+        msg['Subject'] = "Acq Anomaly: Obsid {} (mica processing)".format(anom['obsid'])
+        msg['To'] = ",".join(emails)
+        s = smtplib.SMTP('head.cfa.harvard.edu')
+        s.sendmail('aca@head.cfa.harvard.edu', emails, msg.as_string())
+        s.quit()
+
 def calc_stats(obsid):
     obspar = mica.archive.obspar.get_obspar(obsid, version='last')
     if not obspar:
@@ -643,7 +693,10 @@ def update(opt):
         if not len(acq_stats):
             logger.info("Skipping obsid {}, no stats determined".format(obsid))
             continue
+
         t = table_acq_stats(obsid_info, acq_stats, star_info, catalog, temp)
+        # Check for acquisition anomalies
+        warn_on_acq_anom(t, opt.email)
         _save_acq_stats(t)
 
 
@@ -665,6 +718,8 @@ def get_stats(filter=True):
 
 def main():
     opt = get_options()
+    if opt.email is None:
+        opt.email = DEFAULT_EMAIL
     update(opt)
 
 


jeanconn:
Mica could use actual infrastructure for emailed alerts, but mostly I'm just sick of seeing the bogus acquisition anomaly messages and this should at least start along a path to fixing that (the next step would be to remove the warning from star_stat_db).

mica/stats/acq_stats.py
```diff
@@ -441,6 +448,43 @@ def _get_obsids_to_update(check_missing=False):
     return obsids
 
 
+def warn_on_acq_anom(acq, starcheck, to_addr):
```
taldcroft: Every new "significant" function should have a docstring that describes what it does and the keyword inputs.  Maybe some teeny functions which are completely obvious can have a pass, but as reviewer I spent too much time figuring out just what is `acq`.

mica/stats/acq_stats.py
```diff
@@ -441,6 +448,43 @@ def _get_obsids_to_update(check_missing=False):
...
+    BOX_PAD = 16
+    anom_match = (((np.abs(acq['yang_obs'] - acq['yang']) > (acq['halfw'] + BOX_PAD))
```
taldcroft: Comment like `# Find slots where the star is acquired outside the search box (with padding)`.

mica/stats/acq_stats.py
```diff
@@ -441,6 +448,43 @@ def _get_obsids_to_update(check_missing=False):
...
+                   | (np.abs(acq['zang_obs'] - acq['zang']) > (acq['halfw'] + BOX_PAD))))
+    if not np.any(anom_match):
+        return
+    for anom_idx in np.flatnonzero(anom_match):
+        anom = acq[anom_idx]
+        if anom['img_func'] == "NONE":
+            # if there was nothing tracked, there is really isn't an anomaly
+            continue
+        other_box_match =  ((np.abs(acq['yang'] - anom['yang_obs']) < (acq['halfw'] + BOX_PAD))
```
taldcroft: Comment `# See if the anomalous star was found within a different search box`

mica/stats/acq_stats.py
```diff
@@ -441,6 +448,43 @@ def _get_obsids_to_update(check_missing=False):
...
+def warn_on_acq_anom(acq, starcheck, to_addr):
+    BOX_PAD = 16
+    anom_match = (((np.abs(acq['yang_obs'] - acq['yang']) > (acq['halfw'] + BOX_PAD))
+                   | (np.abs(acq['zang_obs'] - acq['zang']) > (acq['halfw'] + BOX_PAD))))
+    if not np.any(anom_match):
+        return
+    for anom_idx in np.flatnonzero(anom_match):
+        anom = acq[anom_idx]
+        if anom['img_func'] == "NONE":
+            # if there was nothing tracked, there is really isn't an anomaly
```
taldcroft: Comments are generally capitalized.

mica/stats/acq_stats.py
```diff
@@ -441,6 +448,43 @@ def _get_obsids_to_update(check_missing=False):
     return obsids
 
 
+def warn_on_acq_anom(acq, starcheck, to_addr):
+    BOX_PAD = 16
+    anom_match = (((np.abs(acq['yang_obs'] - acq['yang']) > (acq['halfw'] + BOX_PAD))
+                   | (np.abs(acq['zang_obs'] - acq['zang']) > (acq['halfw'] + BOX_PAD))))
+    if not np.any(anom_match):
```
taldcroft: Not necessary (`np.flatnonzero(All False)` gives an empty list).

mica/stats/acq_stats.py
```diff
@@ -441,6 +448,43 @@ def _get_obsids_to_update(check_missing=False):
...
+        if np.any(other_box_match):
+            text = "Acquisition Anomaly.  Star for slot {} actually in box {} \n".format(
+                anom['slot'], acq[other_box_match][0]['slot'])
+        else:
+            text = "Does not appear to be classic star-in-wrong-box anomaly\n"
+        text += """Large Deviation from Expected ACQ Star Position in {obsid}
+      Slot {slot} Expected (Y-Pos, Z-Pos) = ({yang:.1f}, {zang:.1f})
+      Slot {slot} Observed (Y-Pos, Z-Pos) = ({yang_obs:.1f}, {zang_obs:.1f})
+      Halfwidth {halfw:03d}        (dy, dz) = ({dy:.1f}, {dz:.1f})""".format(
+            **{'obsid': anom['obsid'], 'yang': anom['yang'], 'zang': anom['zang'],
```
taldcroft: You can use the more concise `.format(obsid=anom['obsid'], yang=anom['yang'], ..)` or else define a dict before the string format statement for code clarity.

mica/stats/acq_stats.py
taldcroft: It will also help if you adhere to the [recommended idiom](https://occweb.cfa.harvard.edu/twiki/bin/view/Aspect/SkaPython#Variable_and_Method_Names) of using a plural for list-like things where it is not obvious.  For instance `acqs` would give me a hint right away that this is not one acq star but instead a list of table of stars.  Likewise `emails` would hint that is a list of email addresses.

taldcroft: `starcheck` does not appear to be used.

mica/stats/acq_stats.py
jeanconn: For this I was trying to remember if Table.Row had an option to give me a dict() without doing it manually.

jeanconn:
Thanks for review @aldcroft.  Wasn't really done with this yet...  Probably should have put more PR comments indicating that.

jeanconn:
Overall, with regard to eventually deprecating the warnings from the old system, that code warns for acquisition anomalies, 3 or fewer acq stars id'd, and fid drops.  This PR could do the first one.  The kadi StarAcq code is already doing the second one (though I don't know if that will continue in perpetuity), and we probably need to add mica fid stats to support the last one.

mica/stats/acq_stats.py
```diff
@@ -441,6 +448,50 @@ def _get_obsids_to_update(check_missing=False):
...
+    expected search box (plus a pad).  For acquisition stars with tracked positions in the wrong
+    search box, note the box (classic acquisition anomaly).
+
+    :param acqs: astropy table of acquisition stats, includes expected and observed position,
+                 obsid, slot, and halfw used in this method.
+    :param emails: list of addresses to receive email warning if any are generated
+    """
+    # Find tracked objects in the acq stats table outside the intended search box plus padding
+    # img_func == NONE means nothing was tracked
+    box_pad = 16  # arcsecs
```
jeanconn: I don't actually have anything to back up this box pad.

mica/stats/acq_stats.py
```diff
@@ -441,6 +448,50 @@ def _get_obsids_to_update(check_missing=False):
...
+    anom_match = ((acqs['img_func'] != 'NONE') &
+                  ((np.abs(acqs['yang_obs'] - acqs['yang']) >= (acqs['halfw'] + box_pad))
+                   | (np.abs(acqs['zang_obs'] - acqs['zang']) >= (acqs['halfw'] + box_pad))))
+    for anom in acqs[anom_match]:
+        # Check to see if the star is actually found in another box.
+        other_box_match =  ((np.abs(anom['yang_obs'] - acqs['yang']) <= (acqs['halfw'] + box_pad))
+                            & (np.abs(anom['zang_obs'] - acqs['zang']) <= (acqs['halfw'] + box_pad)))
+        if np.any(other_box_match):
+            text = "Acquisition Anomaly.  Star for slot {} actually in box {} \n".format(
+                anom['slot'], acqs[other_box_match][0]['slot'])
```
jeanconn: This will only print the first match if the star would fall into more than one box.   I don't think we care.

mica/stats/acq_stats.py
```diff
@@ -441,6 +448,50 @@ def _get_obsids_to_update(check_missing=False):
...
+        output_dict['dy'] = anom['yang_obs'] - anom['yang']
+        output_dict['dz'] = anom['zang_obs'] - anom['zang']
+        text += """Large Deviation from Expected ACQ Star Position in {obsid}
+      Slot {slot} Expected (Y-Pos, Z-Pos) = ({yang:.1f}, {zang:.1f})
+      Slot {slot} Observed (Y-Pos, Z-Pos) = ({yang_obs:.1f}, {zang_obs:.1f})
+      Halfwidth {halfw:03d}        (dy, dz) = ({dy:.1f}, {dz:.1f})""".format(
+            **output_dict)
+        # Log and Send message for slot.  Obsid can have more than one email
+        logger.warn(text)
+        msg = MIMEText(text)
```
jeanconn: Comparing this borrowed code to the kadi email code, the jobwatch email code etc, wondering if we should put a generic email utility someplace, or if these few lines of smtplib and MIMEText are short "enough".

mica/stats/acq_stats.py
```diff
@@ -441,6 +447,50 @@ def _get_obsids_to_update(check_missing=False):
...
+        msg['From'] = 'aca@head.cfa.harvard.edu'
```
jeanconn: @taldcroft do you prefer these to come from the cron-task USER?  I note that's what you did in the kadi code most recently.

jeanconn:
For testing, I had done some spot checks, but here I've somewhat codified the spot checks.  Note that this isn't quite end-to-end testing.
```
from mica.stats import acq_stats
obsids = [20001, 50445, 18302, 20045]
for obsid in obsids:
    t = acq_stats.table_acq_stats(*acq_stats.calc_stats(obsid))
    acq_stats.warn_on_acq_anom(t, ['jeanconn@head.cfa.harvard.edu'])
```

For these obsids
 - 20001: nominal obsid with no anomaly
 - 50445: obsid with "bogus" anomaly email from star stats processing
 - 18302: obsid that has acquisition outside search box but doesn't appear to be in another box
 - 20045: obsid with two acquisition anomalies

And for that code I received, as expected, no message for 20001 or 50445, 1 message for 18302, and 2 messages for 20045:

```
Subject: Acq Anomaly: Obsid 18302 (mica processing)

Does not appear to be classic star-in-wrong-box anomaly
Large Deviation from Expected ACQ Star Position in 18302
      Slot 5 Expected (Y-Pos, Z-Pos) = (318.0, 513.0)
      Slot 5 Observed (Y-Pos, Z-Pos) = (124.0, 364.8)
      Halfwidth 160        (dy, dz) = (-194.1, -148.2)
```

```
Subject: Acq Anomaly: Obsid 20045 (mica processing)

Acquisition Anomaly.  Star for slot 7 actually in box 1
Large Deviation from Expected ACQ Star Position in 20045
      Slot 7 Expected (Y-Pos, Z-Pos) = (-1599.0, -1225.0)
      Slot 7 Observed (Y-Pos, Z-Pos) = (-1782.8, 1529.3)
      Halfwidth 120        (dy, dz) = (-183.8, 2754.3)
```

```
Acq Anomaly: Obsid 20045 (mica processing)
Acquisition Anomaly.  Star for slot 1 actually in box 7
Large Deviation from Expected ACQ Star Position in 20045
      Slot 1 Expected (Y-Pos, Z-Pos) = (-1814.0, 1505.0)
      Slot 1 Observed (Y-Pos, Z-Pos) = (-1568.5, -1202.9)
      Halfwidth 120        (dy, dz) = (245.5, -2707.9)
```

jeanconn:
@taldcroft I think the only change requested at this point was to show testing.  Does this need a unit test (seems a little overkill to me) or is this good to go?

PR #132, opened 2017-06-01
Add a utility method to get quats from asol ranges

jeanconn:
Add a utility method to get quats from ranges of ground aspect solution files.

### DIFF
--- a/mica/archive/asp_l1.py
+++ b/mica/archive/asp_l1.py
@@ -150,7 +150,7 @@ def get_files(obsid=None, start=None, stop=None,
                              revision=revision, content=content)
 
 
-def get_atts(obsid=None, start=None, stop=None, revision=None):
+def get_atts(obsid=None, start=None, stop=None, revision=None, filter=True):
     """
     Get the ground aspect solution quaternions and times covering obsid or start to stop,
     in the ACA frame.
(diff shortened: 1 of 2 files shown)


jeanconn:
This should get some actual tests in mica.  I thought maybe a utility here would make more sense than including this "quat fetching" in chandra_aca.

mica/archive/asp_l1.py
```diff
@@ -147,6 +150,48 @@ def get_files(obsid=None, start=None, stop=None,
...
+    """
+    Get the ground aspect solution quaternions and times covering obsid or start to stop,
+    in the ACA frame.
+
+    :obsid: obsid
+    :start: start time (DateTime compat)
+    :stop: start time (DateTime compat)
+    :revision: aspect pipeline processing revision (integer version, None, or 'last')
+
+    :returns: Nx4 np.array of quaternions, N times, list of M header records from asol files
```
taldcroft: Say `'list of dict with header from each asol file`.

jeanconn:
I should probably also add a "filter" option to filter on asp_sol_status (quality).

mica/archive/asp_l1.py
```diff
@@ -147,6 +150,48 @@ def get_files(obsid=None, start=None, stop=None,
...
 
 
+def get_atts(obsid=None, start=None, stop=None, revision=None):
+    """
+    Get the ground aspect solution quaternions and times covering obsid or start to stop,
+    in the ACA frame.
+
+    :obsid: obsid
+    :start: start time (DateTime compat)
+    :stop: start time (DateTime compat)
```
jeanconn: typo

jeanconn:
I was thinking about filtering for asp_sol_status,  but I realize that I don't really know what the time stamps in the aqual file represent.  I've got 4 asol entries for each aqual entries, and don't know if the values in aqual represent status on the previous 4 asol samples, a midpoint, or the next 4 samples.

mica/archive/asp_l1.py
```diff
@@ -147,6 +150,48 @@ def get_files(obsid=None, start=None, stop=None,
...
+    :obsid: obsid
+    :start: start time (DateTime compat)
+    :stop: stop time (DateTime compat)
+    :revision: aspect pipeline processing revision (integer version, None, or 'last')
+
+    :returns: Nx4 np.array of quaternions, np.array of N times, list of dict with header from each asol file.
+    """
+    if revision == 'all':
+        raise ValueError("revision 'all' doesn't really make sense for this function")
+    asol_files = sorted(get_files(obsid=obsid, start=start, stop=stop,
```
jeanconn: For the general/time case, these need to be sorted by time and not by file name (which has the obsid in the path).

mica/archive/asp_l1.py
```diff
@@ -147,6 +150,59 @@ def get_files(obsid=None, start=None, stop=None,
...
+        aqual = Table.read(aqual_f)
+        # Check that the time ranges match from the fits headers (meta in the table)
+        if not np.allclose(np.array([asol.meta['TSTART'], asol.meta['TSTOP']]),
+                           np.array([acal.meta['TSTART'], acal.meta['TSTOP']]),
+                           atol=10):
+            raise ValueError("ACAL and ASOL have mismatched time ranges")
+        if filter and np.any(aqual['asp_sol_status'] != 0):
+            # For each sample with bad status find the overlapping time range in asol
+            # and remove from set
+            for idx in np.flatnonzero(aqual['asp_sol_status']):
```
jeanconn: This isn't vectorized, but there should not be many samples with bad status.

jeanconn: Instead of worrying about whether the aqual time covers the previous 4 asol samples or the next 4 samples, here I've just excluded all asol within 4 samples (1.025s).

PR #133, opened 2017-06-09
Use smaller markers on the vv plots

jeanconn:
Use smaller markers on the vv plots .  This is done via a new one-line stylesheet.  

Other options to make the dots smaller included changing every "plot" call to explicitly define markersize (or a new markersize variable) or setting this via rcParams.  The problem with rcParams is that a change will creep into the shared matplotlib namespace, so that seems a side-effect.  Using the a tiny stylesheet localized with a decorator seems like it gives more control without side-effects.

### DIFF
--- a/mica/vv/core.py
+++ b/mica/vv/core.py
@@ -26,6 +26,7 @@ from astropy.table import Table
 from Ska.astro import sph_dist
 from Ska.engarchive import fetch
 import six
+import functools
 
 class NumpyAwareJSONEncoder(json.JSONEncoder):
     def default(self, obj):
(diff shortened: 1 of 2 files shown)


jeanconn:
old
big_dots

new
small_dots

jeanconn:
For this PR, I could probably use an opinion @taldcroft .  I like the smaller dots because I think it makes it easier to see the distribution on the y/z plots.  It also doesn't pile up quite as much on the vs time plots.   

Not a high priority issue, but if I'm going to remake all the reports to address some of the issues brought up when doing some mica data mining for 19387, it would be nice to get this change before remaking all the plots.

jeanconn:
Tom didn't like the small dots on the vs time plots so closing this unmerged.

PR #134, opened 2017-07-02
Smaller markers on just the yz v&v plots

jeanconn:
Use smaller markers on the YZ V&V plots to see the shape of the distribution. 

At some point a "heat map" style density plot for these could also make sense.


jeanconn:
This replaces https://github.com/sot/mica/pull/133 though I'll probably leave that around for a while as an example of how to use a matplotlib stylesheet.

Old output:

PR #135, opened 2017-07-02
Simplify mica.reports (configuration, database access, tables3 compatibility)

jeanconn:
This PR simplifies mica.report by removing some of the "over-configuration", using the mica guide and acq stats databases for the star reports, and adding use of the tables3_api compatibility code.


jeanconn:
Testing:

```
mkdir -p test_report/webreports

In [1]: import mica.report.report
In [2]: mica.report.report.REPORT_ROOT = 'test_report/webreports'
In [3]: mica.report.report.REPORT_SERVER = 'test_report/report_processing.db3'
In [4]: mica.report.report.main(18302)
```

jeanconn:
The other thing on my short list was that I wanted to move the templates into the package (or at least some place easier to test), so I've added that change in this PR as well.

mica/report/report.py
```diff
@@ -472,8 +472,7 @@ def main(obsid):
         os.makedirs(outdir)
 
     jinja_env = jinja2.Environment(
-        loader=jinja2.FileSystemLoader(
-            os.path.join(os.environ['SKA'], 'data', 'mica', 'templates')))
+        loader=jinja2.PackageLoader('mica.report', 'templates'))
```
jeanconn: This doesn't seem to work in the presence of the `mica...egg-info` directory in site-packages.  Not sure if I should dig into the PackageLoader or just switch this to a FileSystemLoader that uses a relative path.

PR #136, opened 2017-07-03
Handle case of obsid with multiple dwells

jeanconn:
Handle case of obsid with multiple dwells.  Obsids 15175 and 4911 have multiple dwells and were skipped in the previous version of this code.    Obsids have been added to test code.

### DIFF
--- a/mica/stats/acq_stats.py
+++ b/mica/stats/acq_stats.py
@@ -451,9 +451,9 @@ def calc_stats(obsid):
     try:
         manvrs = events.manvrs.filter(obsid=obsid, n_dwell__gt=0)
         dwells = events.dwells.filter(obsid=obsid)
-        if manvrs.count() and dwells.count() == 1:
-            manvr = manvrs[0]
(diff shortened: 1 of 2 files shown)


jeanconn:
So, this PR changes the behavior of the acq stats code from totally skipping observations with multiple aspect intervals, to putting the info on the acquisition at the beginning of the first dwell into the database.

As a separate question, do we want to include modifications to the database to be able to store the data from multiple aspect intervals?  In 15175, we have multiple intervals because (I think) the full reacquisition sequence was run, so there may be value in that data.  Alternatively, we could promote kadi StarAcq and use something like that for the weird cases if we want easier access to those results.

taldcroft:
> As a separate question, do we want to include modifications to the database to be able to store the data from multiple aspect intervals? 

Let's wait until a situation arises where we actually need that data.  The re-acqs are a slightly different beast, and we don't have enough to generate very meaningful statistics.  So I don't know what I would do with re-acq data if we had it.

PR #138, opened 2017-07-06
Separate method to get attitudes from asp1 files into new method

jeanconn:
Separate method to get attitudes from asp1 files into new method.

### DIFF
--- a/mica/archive/asp_l1.py
+++ b/mica/archive/asp_l1.py
@@ -159,6 +159,7 @@ def get_atts(obsid=None, start=None, stop=None, revision=None, filter=True):
     :start: start time (DateTime compat)
     :stop: stop time (DateTime compat)
     :revision: aspect pipeline processing revision (integer version, None, or 'last')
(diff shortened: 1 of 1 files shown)


jeanconn:
When working with custom ASP1 processing (in this case, calculating centroid residuals on ERs), I found that I really wanted a method to do that attitude fetching and ACAL transforms on the files that I had in that custom processing directory, not files in the mica archive.  This PR splits that "getting the atts" portion of the code into a separate method for this convenience.

jeanconn:
No new tests were added, but the previous tests on get_atts pass.

jeanconn:
So, @taldcroft per conversation, do you think the `get_atts_from_files` method should stay here or move over to chandra_aca?  I think here still makes the most sense in context.

jeanconn:
I think `get_atts_from_files` could work on any ground aspect solution sets of files, not just files already archived in mica.archive (and that was actually my application).

Issue #142, opened 2017-07-12
Guide stats fetch and MSID alignment should be done with idiomatic code


jeanconn:
See notes from @taldcroft https://github.com/sot/mica/pull/141#issuecomment-314848081

jeanconn:
Closed by #143

PR #143, opened 2017-07-13
Use more idiomatic fetch and filter syntax

jeanconn:
Improve guide stats telem fetching and filtering per recommendation

### DIFF
--- a/mica/stats/guide_stats.py
+++ b/mica/stats/guide_stats.py
@@ -209,23 +209,18 @@ def get_data(start, stop, obsid=None, starcheck=None):
 
     start_time = DateTime(start).secs
     stop_time = DateTime(stop)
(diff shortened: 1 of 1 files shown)


jeanconn:
Closes #142

jeanconn:
Not sure what to do wrt to testing for this one, as I don't have any obsids handy that have bad data to confirm the updated filtering results are as-expected.

taldcroft:
You can at least do a one-off by running this function standalone for any representative obsid and confirm the same results for the current and new code (by e.g. writing `eng_data` to a file).

jeanconn:
I haven't written `eng_data` to a file, but I confirmed for obsid 17861 that the more idiomatic code filters out the same three rows of pcad0 data as the original code and that these rows have msids with bad status/quality.

PR #144, opened 2017-07-17
Use estimated-attitude-corrected box position to check for acquisition anomaly

jeanconn:
Use estimated-attitude-corrected box position to check for acquisition anomaly.  

The idea here is that the acq stats database already has these `dy` and `dz` elements which are calculated as the difference between the expected and actual position of the star, where the expected position is determined via the estimated attitude quaternion.  Therefore the `dy` `dz` of the tracked object is a little bit better at determining how far from the intended position the tracked object was than using "observed - catalog".  However, this isn't *quite* as good as using the actual commanded search box position, which would be the catalog position corrected for the estimated attitude *at the time of the search*.  The `dy` `dz` here in the database are at the time of the transition to guide.  

The acquisition anomaly code already has the padding for dither, however, so this improvement from "observed - catalog" to "dy / dz" looks to be enough to fix the one false-positive the old code demonstrated.

### DIFF
--- a/doc/acq_stats.rst
+++ b/doc/acq_stats.rst
@@ -50,10 +50,10 @@ min_trak_cdz    min "corrected" delta-z during acquisition (arcsec)
 max_trak_mag    max observed magnitude during acquisition
 mean_trak_mag   mean observed magnitude during acquisition
 min_trak_mag    min observed magnitude during acquisition
-cdy             "corrected" delta-y at guide transition (arcsec)
+cdy             "corrected" delta-y at guide transition. (arcsec)
 cdy             "corrected" delta-z at guide transition (arcsec)
-dy              delta-y at guide transition (arcsec)
-dz              delta-z at guide transition (arcsec)
+dy              delta-y at guide transition. expected position determined using onboard est. attitude (arcsec)
+dz              delta-z at guide transition. expected position determined using onboard est. attitude (arcsec)
 ion_rad         ionizing radiation flag set at transition (boolean)
 def_pixel       defective pixel flag set at transition (boolean)
 mult_star       multiple star flag set at transition (boolean)
--- a/doc/acq_stats.rst
+++ b/doc/acq_stats.rst
@@ -50,10 +50,10 @@ min_trak_cdz    min "corrected" delta-z during acquisition (arcsec)
 max_trak_mag    max observed magnitude during acquisition
 mean_trak_mag   mean observed magnitude during acquisition
 min_trak_mag    min observed magnitude during acquisition
-cdy             "corrected" delta-y at guide transition. (arcsec)
-cdy             "corrected" delta-z at guide transition (arcsec)
-dy              delta-y at guide transition. expected position determined using onboard est. attitude (arcsec)
-dz              delta-z at guide transition. expected position determined using onboard est. attitude (arcsec)
+cdy             "corrected" (see below) delta-y at guide transition. (arcsec)
+cdy             "corrected" (see below) delta-z at guide transition (arcsec)
+dy              delta-y at guide transition. expected position (see below) determined using onboard est. attitude (arcsec)
+dz              delta-z at guide transition. expected position (see below) determined using onboard est. attitude (arcsec)
 ion_rad         ionizing radiation flag set at transition (boolean)
 def_pixel       defective pixel flag set at transition (boolean)
 mult_star       multiple star flag set at transition (boolean)
@@ -86,8 +86,20 @@ known_bad       ignore this star in standard processing (boolean)
 bad_comment     reason to ignore a "known_bad" star
 =============== ====================================================================
 
-For the columns that reference "corrected" y-angle and z-angle, these have been
-corrected by the one-shot quaternion update used during the acquisition sequence.
+Note that dy/dz are observed yag/zag (t_guide) - predicted yag/zag (t_guide) using AOATTQT
+(estimated attitude). Observed yag/zag are from AOAC{Y,Z}AN, and t_guide is the time of
+the first sample with AOACASEQ = 'GUID'. t_guide is the same as manvrs.guide_start in
+kadi. The one-shot attitude update occurs in telemetry on the sample after the GUID
+transition.  In particular, dy/dz are used in the check for wrong box acquisition anomaly
+as the the estimated attitude for dy/dz gives a reasonable approximation of the
+OBC estimated attitude at the time of commanding the search boxes. (though the estimated attitude at
+t_acq would be even better for that)
+
+For the columns that reference "corrected" y-angle and z-angle, cdy/cdz, these are the
+observed yag/zag (t_guide) - predicted yag/zag (t_guide) using the estimated attitude at
+that time transformed/corrected by the one-shot quaternion (which is actually calculated from the
+*next* one or two samples).
+
 
 Processing
 ^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


taldcroft:
I guess I'm a little confused, so we need to be really precise with language and times.

It sounds like `dy/dz` are `observed yag/zag (t_guide) - predicted yag/zag (t_guide) using AOATTQT<n>` (estimated attitude).  Observed yag/zag are from `AOAC{Y,Z}AN<n>`, and `t_guide` is the time of the first sample with AOACASEQ = guide (if I have the MSID right).

Now the question is about the estimated attitude at `t_guide`, because I thought that would reflect the one-shot update.  For instance imagine a maneuver with no attitude disturbances so that the control law does its job and at the end of the maneuver (and start of acquisition), the estimated attitude is exactly equal to both the target and commanded attitudes.  (For the sake of argument set dither=0).  However, in truth there is a 100 arcsec attitude error (but the OBC does not yet know that).  So the acquisition search boxes will be at precisely the catalog values because the on-board estimate of attitude error and dither are both 0.

Now if you use the attitude estimate at `t_guide` to infer where the OBC commanded the boxes, you will be off by 100 arcsec.  However, if you use the OBC attitude estimate just before the one-shot update then that will be closer.

jeanconn:
> I guess I'm a little confused, so we need to be really precise with language and times.

Indeed, and if you see places where I should improve docs or process in here, please let me know.  The fundamental problem is that my vocabulary was evolving around these issues as the acq_stats code was being written.

> It sounds like dy/dz are observed yag/zag (t_guide) - predicted yag/zag (t_guide) using AOATTQT<n> (estimated attitude). Observed yag/zag are from AOAC{Y,Z}AN<n>, and t_guide is the time of the first sample with AOACASEQ = guide (if I have the MSID right).

Yes. 

> Now the question is about the estimated attitude at t_guide, because I thought that would reflect the one-shot update.

That's where I disagree.  The estimated attitude does not get updated for the one-shot update until after t_guide.

doc/acq_stats.rst
```diff
@@ -50,10 +50,10 @@ min_trak_cdz    min "corrected" delta-z during acquisition (arcsec)
 max_trak_mag    max observed magnitude during acquisition
 mean_trak_mag   mean observed magnitude during acquisition
 min_trak_mag    min observed magnitude during acquisition
-cdy             "corrected" delta-y at guide transition (arcsec)
+cdy             "corrected" delta-y at guide transition. (arcsec)
 cdy             "corrected" delta-z at guide transition (arcsec)
-dy              delta-y at guide transition (arcsec)
-dz              delta-z at guide transition (arcsec)
+dy              delta-y at guide transition. expected position determined using onboard est. attitude (arcsec)
+dz              delta-z at guide transition. expected position determined using onboard est. attitude (arcsec)
```
jeanconn: Maybe this should read "delta-z at guide transition (before one-shot update).  expected position determined using onboard est. attitude (arcsec)"

taldcroft:
> That's where I disagree. The estimated attitude does not get updated for the one-shot update until after t_guide.

OK, I looked at one case and agree that the attitude is updated on the *next* telemetry sample after AOACASEQ goes from AQXN to GUID.  It seems like a bit of a potential race condition to count on that, but I guess in practice it works and is convenient.

jeanconn:
In acq_stats, it looks like it has been OK thus far to use the estimated atttitude at t_guide and the estimated attitude from the sample *after* the next sample to calculate a delta quaternion and call that the one-shot-update quaternion.  

This was before I was as comfortable with the attitude errors just being called the one-shot.

doc/acq_stats.rst
jeanconn: Right, though for "reasonable approximation" that is the point I was making in the opening of the PR that we'd be even better off using the estimated attitude at the time of the transition to AQXN, but that isn't already convenient in the database.

jeanconn:
From the anomaly warning on 19586, it looks like the img_func can be non-NONE and still have nothing tracked with reasonable position.  I should probably try to fix that in this PR as well.
```
Does not appear to be classic star-in-wrong-box anomaly
Large Deviation from Expected ACQ Star Position in 19586
      Slot 2 Expected (Y-Pos, Z-Pos) = (-2212.0, -1352.0)
      Slot 2 Observed (Y-Pos, Z-Pos) = (0.0, 0.0)
      Halfwidth 120        (dy, dz) = (2212.0, 1352.0)
```

mica/stats/acq_stats.py
```diff
@@ -461,9 +461,9 @@ def warn_on_acq_anom(acqs, emails):
     # Find tracked objects in the acq stats table outside the intended search box plus padding
     # img_func == NONE means nothing was tracked
     box_pad = 16  # arcsecs
-    anom_match = ((acqs['img_func'] != 'NONE') &
-                  ((np.abs(acqs['yang_obs'] - acqs['yang']) >= (acqs['halfw'] + box_pad))
-                   | (np.abs(acqs['zang_obs'] - acqs['zang']) >= (acqs['halfw'] + box_pad))))
+    anom_match = ((acqs['img_func'] != 'NONE') & (acqs['img_func'] != 'SRCH') &
```
jeanconn: Adding "SRCH" to the filter here stops the bogus warnings for 19586 19814 etc.

mica/stats/acq_stats.py
```diff
@@ -475,12 +469,15 @@ def warn_on_acq_anom(acqs, emails):
             text = "Does not appear to be classic star-in-wrong-box anomaly\n"
         # Make a dictionary of the anom record for use in string formatting
         output_dict = {col: anom[col] for col in anom.colnames}
-        output_dict['dy'] = anom['yang_obs'] - anom['yang']
-        output_dict['dz'] = anom['zang_obs'] - anom['zang']
+        output_dict['dy'] = anom['dy']
+        output_dict['dz'] = anom['dz']
         text += """Large Deviation from Expected ACQ Star Position in {obsid}
-      Slot {slot} Expected (Y-Pos, Z-Pos) = ({yang:.1f}, {zang:.1f})
+      Slot {slot} Expected* (Y-Pos, Z-Pos) = ({yang:.1f}, {zang:.1f})
```
taldcroft: Is that `*` a "fat-finger"?  (Brent's word).

jeanconn: I really was trying for a "see the asterisk note below", so it isn't a "fat-finger", but it looks like it wasn't effective.

taldcroft: Yeah, just skip the asterisk, it is confusing.

mica/version.py
```diff
@@ -14,7 +14,7 @@
 therein are applicable.
 """
 
-version = '3.12.1'
+version = '3.12.2'
```
jeanconn: OK. I think this PR is reasonable as a bugfix so I've incremented the version accordingly.

taldcroft:
Just one more comment.  I counted the number of acq anomalies this reporting period for the Quarterly (by looking at alert emails) and got 12.  It might be nice to have the list of acq anomalies automatically updated, or have a status flag, or something.

50157
18450
20079
19700
50045
20095
19598
19659
49835
49817
19817
20795

jeanconn:
My thought was to have the list and details on acq anomalies in separate
new reporting code,  but I'm not sure what you have in mind with this
comment on this pr.

taldcroft:
I was just thinking that the current code here has all we need to know about an acq anomaly, so it's a good place to just drop that info somewhere.  In other words, the numbers here should be stored somewhere that is easy to get, and then we're done:
```
Acquisition Anomaly.  Star for slot 1 actually in box 6
Large Deviation from Expected ACQ Star Position in 20795
      Slot 1 Expected (Y-Pos, Z-Pos) = (571.0, 1687.0)
      Slot 1 Observed (Y-Pos, Z-Pos) = (-688.4, -1062.8)
      Halfwidth 120        (dy, dz) = (-1259.3, -2749.8)
```
But in the quest for good rather than perfect (and smaller PRs), this PR is good enough and can be merged and installed.  I'm opening another issue.

Issue #137, opened 2017-07-03
mica.starcheck identifies no-starcat obsids as being "pre timelines"


jeanconn:
For example, obsid 49961 (a dark cal replica) has no star catalog, so it has no value for "mp_starcat_time" in the database of starchecks.  None is apparently less than '2001:001' etcs in a date comparison at:

https://github.com/sot/mica/blob/32a3291db98d14c638080c610fe5d612147ffdaf/mica/starcheck/starcheck.py#L219

PR #146, opened 2017-07-18
Fix inconsistencies in queried starcheck database for observations without star catalogs

jeanconn:
Fix inconsistencies in queried starcheck database for observations without star catalogs.  The fixes include:

- ingesting maneuver information for observations with maneuver but without star catalog in starcheck output
- explicitly handling the case of an observation with a None/NULL entry for the star catalog time


jeanconn:
Closes #137

PR #147, opened 2017-07-18
Fix update() bug introduced by ec6f166

jeanconn:
The update to fetch cmd_states from fetch_states instead of via the Sybase table introduced a tiny bug/IndexError.  

The update() code is not directly tested by module tests.

### DIFF
--- a/mica/report/report.py
+++ b/mica/report/report.py
@@ -843,7 +843,7 @@ def update():
 
     recent_obs = np.unique(fetch_states(start=DateTime(-7), vals=['obsid'])['obsid'])
     for obs in recent_obs:
-        process_obsids([obs['obsid']])
+        process_obsids([obs])
 
 
 


taldcroft:
Is there an opportunity here to improve testing?

jeanconn:
Define "opportunity" :smirk:

This last round added at least a report-making test.  I'm not sure if it is worth the work right now to extend that to a full mica report update/process test at this point.  To test "update()" directly, would I monkey-patch `fetch_states` to have some fixed values to test?  Or just confirm that an "update" would run at time "now" on an on-the side mica report area and database?  That might be a little long for a module test as well.

jeanconn:
Do we have a convention for a pytest setup to skip long tests?

jeanconn:
It looks like if I just run update() as a test, today that takes an additional 500 seconds.

jeanconn:
I just borrowed some code to skip *unless* --longrun supplied to py.test.  Would something like that make sense here?  This is somewhat self-documenting to remind the developer to run the long tests, though I haven't figured out the best way to get slightly better text
```
1 tests deselected by "-m 'not longrun'" 
```

jeanconn:
I just reverted the "long" test from the module tests and opened an issue #148

mica/report/report.py
```diff
@@ -843,7 +843,7 @@ def update():
 
     recent_obs = np.unique(fetch_states(start=DateTime(-7), vals=['obsid'])['obsid'])
     for obs in recent_obs:
-        process_obsids([obs['obsid']])
+        process_obsids([obs])
```
jeanconn: I'm also not entirely clear on why this routine didn't originally pass the list of obsids to process_obsids and have this be a single function call.  But probably better to do the minimal change to make this run for now and improve it after I add regression testing.

PR #157, opened 2017-09-21
Add LICENSE and license comments

taldcroft:


mica/starcheck/starcheck_parser.py
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env
```
taldcroft: Was this supposed to be `#!/usr/bin/env python`?

PR #159, opened 2017-10-22
Fix bug in parsing/ingest of manvrs

jeanconn:
Code that was intended to support not-breaking on intermediate maneuvers has a regex bug that is causing extra empty maneuvers to get in to the data structure.  This is breaking ingest.  This small bugfix doesn't fix the regex, but just stops the empty manvr entries from ending up in the data structure.


jeanconn:
All the bad parsing code was meant to be somewhat stop-gap until we set up starcheck to put out JSON, but it is persisting.  This is somewhat time-critical as the bug has stopped nominal processing.  If @taldcroft has a chance to review that is great, otherwise I see this as low impact and could go without review.  For testing I've just confirmed that the manvr structure created looks good on recent schedules and an on-the-side run of the OCT1617A ingest works without error.

PR #164, opened 2017-11-21
Only run report-making test if Sybase/OCAT available

jeanconn:


mica/report/tests/test_write_report.py
```diff
@@ -2,9 +2,20 @@
...
 import os
 import shutil
+import pytest
 
 from .. import report
 
+HAS_SYBASE_ACCESS = False
+try:
+    import Ska.DBI
+    with Ska.DBI.DBI(server='sqlsao', dbi='sybase', user='aca_ops', database='axafocat') as db:
```
jeanconn: I think making the actual db connection makes the most sense in this case.  If the tests are run from Py3, from GRETA, or as a user who doesn't have read access on the database, it seems reasonable to just skip the report-making test.

mica/report/tests/test_write_report.py
```diff
@@ -2,9 +2,20 @@
 import tempfile
 import os
 import shutil
+import pytest
 
 from .. import report
 
+HAS_SYBASE_ACCESS = False
```
taldcroft: Just a nit that this should be in the `except` block.  Essentially you are doing an `if / else`, and it makes the code logic more clear.

Yes:
```
if something:
    x = True
else:
    x = False
```
No, because the `x = False` is not explicitly linked to `something` in the code syntax.
```
x = False
if something:
    x = True
```
I'm sure I have written code along the "No" style, but I worry more about code clarity these days.

PR #167, opened 2017-12-13
Move "flight" mica archive into /proj/sot/ska/data/mica/archive

jeanconn:
Change the default location of the flight mica archive.

### DIFF
--- a/mica/archive/archive.conf
+++ b/mica/archive/archive.conf
@@ -1,8 +1,0 @@
-[asp_l1_defaults]
-data_root = asp_l1
-temp_root = temp
-
-[aca_l0_defaults]
-data_root = /data/aca/archive
-temp_root = /data/aca/archive/temp
-content = acadata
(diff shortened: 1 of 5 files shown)


jeanconn:
@taldcroft Maybe this would be the cleanest way to support GRETA?  On the HEAD side it would just be a symbolic link to /data/aca/archive .

jeanconn:
Yes, I probably should have done that remove as a separate commit to comment on that.  As far as I can tell (grepping, and running the module tests), we'd already deprecated archive.conf when we added mica.common .  The only thing that is a bit tricky is that I can't make /data/aca go away easily to test on the HEAD LAN, so I'll want to double-check this on my VM without that mount.

taldcroft:
Can you just check on GRETA?

jeanconn:
Do we want to copy over the minimal set of data to make the L0, L1, and obspar lookup tests pass on GRETA?  I hadn't gotten that far in my thinking (was configuring to make the dark cal work, not the tests pass).

taldcroft:
Passing tests would be nice.  :smile:

jeanconn:
You know I love everything to be totally consistent, so I was a little bit hesitant to, say, copy over the sqlite tables that have the records of all of the asp1, l0, and obspar files... without actually copying over all the files!  It might be valuable to add some nicer handling in mica for missing files (that should be there according to its database) when called on GRETA.

PR #168, opened 2017-12-22
Skip tests based on existence of parts of mica archive

jeanconn:
A first pass at just skipping tests if components of the mica archive are not available to do those tests.

This is ugly due to overconfiguration of the paths.  Will clean that up in future PRs.

I haven't found a good way to skip all of the remaining tests in a file if a condition is met, so just copied the decorators down these files.

### DIFF
--- a/mica/stats/tests/test_guide_stats.py
+++ b/mica/stats/tests/test_guide_stats.py
@@ -1,19 +1,35 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import tempfile
 import os
+import numpy as np
+import pytest
 
 from .. import guide_stats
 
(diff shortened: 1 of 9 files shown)


mica/stats/tests/test_guide_stats.py
```diff
@@ -1,19 +1,35 @@
...
 import os
+import numpy as np
+import pytest
 
 from .. import guide_stats
 
+HAS_GS_TABLE = os.path.exists(os.path.abspath(guide_stats.TABLE_FILE))
 
+@pytest.mark.skipif('not HAS_GS_TABLE', reason='Test requires guide stats table')
+def test_read_stats():
```
jeanconn: This is a new "read" test.

PR #173, opened 2018-05-27
Refactor stats into update and get_stats modules

taldcroft:
This was driven by missing modules in my local Ska which prevented importing `get_stats`. 

Because of the renaming etc, the diffs from master are hard to follow, but basically all that happened was that the `guide_stats.py` => `update_guide_stats.py`, and `get_stats()` was put back into `guide_stats.py` as standalone.  The only other difference is that SKA defaults to /proj/sot/ska.  It turns out that `kadi.events` is actually *setting* the SKA env var if it isn't defined, and that was allowing the `guide_stats.py` module to import without SKA being defined in the shell.  (This side-effect of importing `kadi.events` seems not ideal).

Testing:
```
In [1]: from mica.stats import guide_stats

In [2]: dat = guide_stats.get_stats()

In [3]: from mica.stats import acq_stats

In [4]: dat = acq_stats.get_stats()

In [5]: guide_stats.__file__
Out[5]: '/Users/aldcroft/git/mica/mica/stats/guide_stats.py'
```
I don't know how to test the `update` scripts.


taldcroft:
BTW if you look at the first two commits individually the changes should make sense.  I purposely committed in this way to allow review.

jeanconn:
The changes looked good to me here, but this isn't passing tests, so isn't ready to go.

taldcroft:
Rats.  If I have time I'll look into it, otherwise push it off.

jeanconn:
This rebased cleanly, so I did.

taldcroft:
What did you did?  Fix the tests?

jeanconn:
Haven't done anything yet.  Just rebased to get rid of the merge conflict.  My plan was to test or update tests and get this merged before trying to move the stats jobs to ska3 processing and before adding the get_star_stats() method I'd like in both guide and acq stats.

PR #177, opened 2018-07-22
Add logic for format errors on supplied mp_dir to get_starcheck_catalog

jeanconn:
The expected form for "mp_dir" for get_starcheck catalog is silly (like /2018/APR2318/oflsa/) as it isn't really a directory.  This PR reformats variant of mp_dir so they'll still work with the database query.  "/data/mpcrit1/mplogs/2018/APR2318/oflsa" will work, as will "2018/APR2318/oflsa" .

### DIFF
--- a/mica/starcheck/starcheck.py
+++ b/mica/starcheck/starcheck.py
@@ -310,7 +310,7 @@ def get_mp_dir(obsid, starcheck_db=None, timelines_db=None):
     raise ValueError("get_mp_dir should find something or nothing")
 
 
-def get_starcheck_catalog(obsid, mp_dir=None, starcheck_db=None, timelines_db=None):
+def get_starcheck_catalog(obsid, mp_dir=None, starcheck_db=None, timelines_db=None, mp_top_level=None):
     """
     For a given obsid, return a dictionary describing the starcheck catalog that should apply.
     The content of that dictionary is from the database tables of that parsed the starcheck report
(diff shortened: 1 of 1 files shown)


jeanconn:
Should add a couple of tests.

taldcroft:
Is the new `mp_top_level` parameter needed?  I think supporting these options for `mp_dir` would be sufficient:
- `/[12]\d\d\d/<whatever>`: append to the config `mp_top_level` dir.
- Normal absolute or relative path

So this preserves the existing ("silly") API while adding path specifications that are like every other path spec in linux.  There is basically zero chance that an actual absolute dir path will start with something that looks like a year.

jeanconn:
> Is the new mp_top_level parameter needed?

Not really, I just didn't want to add code that would break the "idea" that this is all relocatable.  It isn't really relocatable.

jeanconn:
Regarding:

> /[12]\d\d\d/<whatever>: append to the config mp_top_level dir.

I don't know what you mean by "append to the config mp_top_level dir" in this context.  The whole problem is that the database has "/dddd/mmmddyy/ofls?/' as the string to match, so the code is stripping off mp_top_level dir to get the string that we want to match.  It doesn't matter if we construct the string we want to match here or one block down.

We could also rebuild the database with full paths or reconstruct the query to use "like" or something else.

jeanconn:
Or I could just leave this alone and let the docstring make the user do what the database query wants.

jeanconn:
Or I suppose we could just look for "\/\d{4}\/\w{3}\d{4}\/\w{5}\/" (or whatever) somewhere in whatever is provided for mp_dir and hope for the best.

jeanconn:
OK, so I brought this PR, because when I was looking at my suggestions to you that you should supply mp_dir to get_starcheck_catalog for the regression weeks for proseco... I tried to supply a full directory for mp_dir and it didn't work.  Because mp_dir isn't a dir.  So, if we can both remember not to supply a dir for this thing called mp_dir, this PR isn't needed.

taldcroft:
At least my problem there was not remembering the format (which is clearly stated in the docs) but entirely forgetting to use `mp_dir`.  If there had been a parameter `load_name` that might have triggered the thought that I need to supply the load name, but maybe not.  I think this is OK for now, so closing.

PR #178, opened 2018-07-24
Fixes to starcheck for specifying mp_dir

### LINKED ISSUE #175: Inconsistent results / exceptions in starcheck queries
This original issue was user error, but subsequent discussion shows inconsistencies in the way that queries for not-available starcheck products are handled.  Would be good to clean up, and in particular raise an exception any time a valid query result cannot be returned.
```
In [28]: get_starcat(21082)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/data/baffin/tom/git/proseco/proseco/tests/starcheck_proseco.py in <module>()
----> 1 get_starcat(21082)

/proj/sot/ska3/flight/arch/x86_64-linux_CentOS-5/lib/python3.6/site-packages/mica/starcheck/starcheck.py in get_starcat(obsid)
     34     """
     35     if obsid not in OBS_CACHE:
---> 36         OBS_CACHE[obsid] = get_starcheck_catalog(obsid)
     37     return OBS_CACHE[obsid]['cat']
     38 

/proj/sot/ska3/flight/arch/x86_64-linux_CentOS-5/lib/python3.6/site-packages/mica/starcheck/starcheck.py in get_starcheck_catalog(obsid, mp_dir, starcheck_db, timelines_db)
    346     status = None
    347     if mp_dir is None:
--> 348         mp_dir, status, obs_date = get_mp_dir(obsid, starcheck_db=starcheck_db, timelines_db=timelines_db)
    349     # if it is still none, there's nothing to try here
    350     if mp_dir is None:

/proj/sot/ska3/flight/arch/x86_64-linux_CentOS-5/lib/python3.6/site-packages/mica/starcheck/starcheck.py in get_mp_dir(obsid, starcheck_db, timelines_db)
    308         else:
    309             return (sc['dir'], 'approved', sc_date)
--> 310     raise ValueError("get_mp_dir should find something or nothing")
    311 
    312 

ValueError: get_mp_dir should find something or nothing
```

taldcroft:
Fixes #175.

To do:

- [x] Tests

### DIFF
--- a/mica/starcheck/starcheck.py
+++ b/mica/starcheck/starcheck.py
@@ -336,6 +336,9 @@ def get_starcheck_catalog(obsid, mp_dir=None, starcheck_db=None, timelines_db=No
     :param timelines_db: optional handle to already-open timelines database
     :returns: dictionary with starcheck content described above
     """
+    # Keep the original kwarg from user for the caching key later
+    mp_dir_kwarg = mp_dir
+
     if (obsid, mp_dir) in OBS_CACHE:
         return OBS_CACHE[obsid, mp_dir]
 
@@ -358,7 +361,6 @@ def get_starcheck_catalog(obsid, mp_dir=None, starcheck_db=None, timelines_db=No
         load_year = '20' + mp_dir[5:7]
         load_name = mp_dir[:7]
         mp_dir = '/{}/{}/ofls{}/'.format(load_year, load_name, load_version)
-        print(mp_dir)
 
     db = starcheck_db # shorthand for the rest of the routine
     sc_id = db.fetchone("select id from starcheck_id "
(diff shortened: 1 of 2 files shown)


jeanconn:
Ah.  I'm now guessing you were using the convenience methods for your proseco testing.
These changes look good to me; thanks for following through with it.

IIRC, the get_mp_dir calls are the parts that are slow as molasses, so when doing lookups and supplying the mp_dir string, the cache may not really be necessary.  But it looks to me like at least it won't be broken.

taldcroft:
> so when doing lookups and supplying the mp_dir string, the cache may not really be necessary.

Can you explain?  I would have thought that the caching will prevent subsequent calls to `get_mp_dir()`.  I must be misunderstanding something.

jeanconn:
I can try to explain.

Yes, if you are using get_starcheck_catalog without supplying an mp_dir, the cache is going to prevent subsequent calls to get_mp_dir.  If you are calling get_starcheck_catalog and supplying mp_dir as an option, there is no first or subsequent call to get_mp_dir.

Since get_mp_dir is the slow thing, the cache doesn't help that much in the case when you are not actually doing the slow thing.  That was all I was trying to say.

taldcroft:
OK, makes sense that it isn't preventing the *really* slow part, but still helps by not spinning up two DBIs and doing sqilte queries, etc.

taldcroft:
Ready for review

PR #180, opened 2018-08-13
Update common.py

taldcroft:
Mica is not happy by default in my standalone environment, where SKA is not /proj/sot/ska.  There is a workaround setting MICA_ARCHIVE, but that shouldn't be necessary in general.


mica/common.py
```diff
@@ -8,7 +8,7 @@
 class MissingDataError(Exception):
     pass
 
-FLIGHT_MICA_ARCHIVE = '/proj/sot/ska/data/mica/archive'
+FLIGHT_MICA_ARCHIVE = os.path.join(os.environ['SKA'], 'data', 'mica', 'archive')
```
jeanconn: I thought the whole point of the FLIGHT_MICA_ARCHIVE was that it was always actually the production mica archive, though I don't know if it matters.  Would it make just as much sense to set to None if not defined on your system (seems truer to the definition)?

taldcroft: I'm not sure what is the point of hardwiring a "production" archive that does not exist on my machine.  And if I'm working in a non-production ska3 and have a dev version of the mica archive for in $SKA/data/mica, then the expected behavior is that it should pick up my local version by default.  It would be surprising to get the /proj/sot/ska version.

I guess I don't care about the internals, but  I do want to be able to get a starcheck catalog or dark cal without defining `MICA_ARCHIVE` in my bash profile.

jeanconn: Sure.  When we set this up, the mica archive was basically the L0 image archive and the L1 solutions archive, we wouldn't plan to have those on a laptop, and the only reason for a non-production archive was for testing.  Use cases have changed.

jeanconn: I'm not sure if if we should split the two, in fact, where the L0 archive and L1 archive stay in MICA_ARCHIVE but dark cal and starcheck go back into SKA/data/mica or the like.

jeanconn: Also did you just delete the /proj/sot/ska/data/mica/archive link?  My annie second annie sim just died and now can't find dark cal data?

taldcroft: In the unlikely event that it is necessary to have split access to different elements of the archive, it seems like using soft links from within $SKA/data/mica would do the trick.

jeanconn: I was mostly thinking about setting up tests.  Right now we end up with different hacks to say "do you really have the relevant piece of the mica archive for this test" when running some tests, when really we mostly care about "do you have the big file mica archive" or do you have the "little useful mica archive" or do you have neither?  But that doesn't need to be an issue for this PR.  Just something to think about as an future enhancement.

jeanconn: Should this be using ska_path for this instead of os.environ['SKA'] ?

taldcroft: Doesn't really matter.  I don't like the current release functionality in ska_path.  I see that https://github.com/sot/ska_path/pull/4 never got merged.  With that, then the only advantage of using ska_path is a better error message if you don't have SKA defined.  

But right now without that PR merged (for version 4.2) then I don't want ska_path.

jeanconn: I just haven't even been setting SKA in my environments because the modules have been correctly getting defaults.

taldcroft: There is no longer a unique default in Ska3 and SKA is a required env var for our core packages, so these packages should no longer assume `/proj/sot/ska` anywhere.  We will eventually clean things up so that things will fail for you without SKA.  :smile:

taldcroft: BTW, I purposely moved data on my standalone platforms out of /proj/sot/ska to identify problems (or rather things that accidentally work) related to this legacy default.

jeanconn: Right and this is why I also wanted to ask @mbaski  about this in the [other thread](https://github.com/sot/proseco/issues/45) so that we wouldn't implement changes that are contrary to what is already assumed by setup scripts on the Window side.

jeanconn:
Seems reasonable to fix the Windows issue in this PR now too.  I don't recall when MICA_ARCHIVE_PATH is expected to have multiple entries, so I don't know if we want to cut that functionality or just do the minimal and fix the incorrect split on colons.

With regard to the split, I don't immediately see a `pathlib` option for splitting a path list, but if we want to keep the last line, string splitting on os.pathsep seems like it would at least be correct by os.

mbaski:
That change (in ae00c60) works great on Windows!

taldcroft:
👍 on `os.pathsep`.  Perfect solution.

jeanconn:
I think this two-line PR is good to go and that once it is in master we're done with a short-term mica release.  Agree @taldcroft ?

taldcroft:
Agreed

jeanconn:
And I think the changes are small and useful in mica so should just go into ska2 as well.  So planning to tag this as a 3.15 release and install in ska2 and ska3.

taldcroft:
Agreed.

PR #186, opened 2019-01-22
Update dither parsing for starcheck 13.0

jeanconn:
Update dither parsing for starcheck 13.0 .

Looks like the dither printing for starcheck 13.0 added an extra space that now needs a small mod to the parsing regex.

### DIFF
--- a/mica/starcheck/starcheck_parser.py
+++ b/mica/starcheck/starcheck_parser.py
@@ -107,7 +107,7 @@ def get_dither(obs_text):
         raise ValueError("No OBSID found for this catalog")
     for oline in targ_search.group(1).split("\n"):
         dither_re = re.match(
-            "Dither:\s(\S+)\s+Y_amp=\s*(\S+)\s+Z_amp=\s*(\S+)\s+Y_period=\s*(\S+)\s+Z_period=\s*(\S+)\s*",
+            "Dither:\s+(\S+)\s+Y_amp=\s*(\S+)\s+Z_amp=\s*(\S+)\s+Y_period=\s*(\S+)\s+Z_period=\s*(\S+)\s*",
             oline)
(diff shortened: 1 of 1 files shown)


jeanconn:
I let the space go in starcheck 13.0 review of regression outputs because I figured the starcheck_parser would need more changes anyway, but it looks like this might be it.

mica/starcheck/starcheck_parser.py
```diff
@@ -107,7 +107,7 @@ def get_dither(obs_text):
         raise ValueError("No OBSID found for this catalog")
     for oline in targ_search.group(1).split("\n"):
         dither_re = re.match(
-            "Dither:\s(\S+)\s+Y_amp=\s*(\S+)\s+Z_amp=\s*(\S+)\s+Y_period=\s*(\S+)\s+Z_period=\s*(\S+)\s*",
+            "Dither:\s+(\S+)\s+Y_amp=\s*(\S+)\s+Z_amp=\s*(\S+)\s+Y_period=\s*(\S+)\s+Z_period=\s*(\S+)\s*",
```
jeanconn: I'll try to just get in a whitespace fix in starcheck so we can skip the change, but wanted this ready.

mica/starcheck/starcheck_parser.py
```diff
@@ -116,7 +116,7 @@ def get_dither(obs_text):
                         dither_y_period=float(dither_re.group(4)),
                         dither_z_period=float(dither_re.group(5)))
         dis_dither_re = re.match(
-            "Dither:\sOFF\s*",
+            "Dither:\s+OFF\s*",
```
jeanconn: Actually, looks like just the ON has a new space so this isn't needed.

PR #188, opened 2019-02-12
Add utility script to create initial agasc_supplement.h5

taldcroft:
The idea is to install the new file `agasc_supplement.h5`  to `$SKA/data/agasc`.  

Later:
- Add some metadata with a key to the `source` column (maybe another table with the same `source` integer key and then a text string).
- Add new tables, e.g.:
  - `observed` : observed mag / mag error, number of times as guide / acq, number of acq fails)
  - `position` : position update

Then there will be a mica update script to keep it up to date on a daily basis.  For now this is useful and defines the basic structure and plan.

To do:
- [x] Add this file to the ska_sync default config

### DIFF
--- a/utilities/make_agasc_supplement.py
+++ b/utilities/make_agasc_supplement.py
@@ -1,0 +1,49 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+"""
+Create the initial version of agasc_supplement.h5.
+
+This file is a supplement to the stable AGASC to inform star selection
+and star catalog checking.
+
+This script simply creates the initial file that has only bad
+stars from two sources:
+ - starcheck bad star list
+   https://github.com/sot/starcheck/blob/master/starcheck/data/agasc.bad
+ - GAIA high proper motion file $SKA/analysis/gaia/agasc_gaia_xmatch_PM_gt_50mas.fits.gz
+   See: https://nbviewer.jupyter.org/url/cxc.cfa.harvard.edu/mta/ASPECT/ipynb/star_selection/gaia
+        GAIA guide star crossmatch.ipynb
+
+"""
+import os
+from pathlib import Path
+
+import numpy as np
+
+from astropy.table import Table
+
+HOME = Path(os.environ['HOME'])
+SKA = Path(os.environ['SKA'])
+
+agasc_ids = []
+sources = []
+
+# Starcheck bad star list is not installed anywhere so just grab from local git repo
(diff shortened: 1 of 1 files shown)


utilities/make_agasc_supplement.py
```diff
@@ -0,0 +1,49 @@
...
+
+from astropy.table import Table
+
+HOME = Path(os.environ['HOME'])
+SKA = Path(os.environ['SKA'])
+
+agasc_ids = []
+sources = []
+
+# Starcheck bad star list is not installed anywhere so just grab from local git repo
```
jeanconn: Well, it is installed in package data but..

utilities/make_agasc_supplement.py
```diff
@@ -0,0 +1,49 @@
...
+lines = open(HOME / 'git' / 'starcheck' / 'starcheck' / 'data' / 'agasc.bad', 'r').readlines()
```
jeanconn: I'd just argue that this somewhat contaminates the new file with this list that is made half of manually added stars and half from star stats.

utilities/make_agasc_supplement.py
taldcroft: In ska3-flight I cannot `import starcheck`, so that file being in the package data doesn't help me too much.

jeanconn: Right, in ska3-flight alone there is no starcheck.

jeanconn:
I'm assuming that whatever h5 is made via astropy.table that everything will work reasonably with pytables in ska3 (was just adding indexes broken in our version)?

utilities/make_agasc_supplement.py
taldcroft: I'm not very concerned about provenance and perfection at this point.  I just consider `agasc.bad` as a legacy source of bad stars with some recent manual additions and a 4-year old list of auto additions.  It's a starting point.

Going forward there will be a new process which takes some ideas from bad_stars.py but is mostly new.

jeanconn: That's fine.  I just thought it would be as useful to import from the manual star file in that same directory and then add the rest of the stars that we're using in proseco / agasc.bad with a new source identifier.  But if you don't care, you don't care.

taldcroft:
Right, this doesn't need its own release and I guess will just get pulled along whenever the next mica release happens.

taldcroft:
This new file needs to get installed on GRETA.  I will do so now.

taldcroft:
Can I remove $SKA/data/agasc/starcat_supplement_1p7.h5?  Or.. what is that?

jeanconn:
That was another first draft; it can go, no problem.

PR #189, opened 2019-02-17
Clip negative dark map values by default

taldcroft:
Slightly controversial, but this makes the default in mica to clip negative values from dark maps returned by `dark_cal` functions.  

Since those values are not physical, I can't think of a good reason to not do this clipping globally by default.  Looking at SOT uses of [get_dark_cal_image](https://github.com/search?p=2&q=org%3Asot+get_dark_cal_image&type=Code) and [get_dark_cal_props](https://github.com/search?q=org%3Asot+get_dark_cal_props&type=Code) doesn't show anything that would break from not having negative pixel values (from what I saw).

There are doubtless non-versioned scripts or notebooks that show dark current distributions that would now be clipped and might require update.

### DIFF
--- a/mica/archive/aca_dark/dark_cal.py
+++ b/mica/archive/aca_dark/dark_cal.py
@@ -109,7 +109,8 @@ def get_dark_cal_id(date, select='before'):
 
 
 @DARK_CAL.cache
-def _get_dark_cal_image_props(date, select='before', t_ccd_ref=None, aca_image=False):
+def _get_dark_cal_image_props(date, select='before', t_ccd_ref=None, aca_image=False,
+                              allow_negative=False):
     """
     Return the dark calibration image (e-/s) nearest to ``date`` and the corresponding
     dark_props file.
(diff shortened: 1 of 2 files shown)


taldcroft:
Why not for Tuesday?

jeanconn:
I mean, I think you are completely right that there nothing that's really dependent on getting the negative values (our dark current reporting isn't using this code, for example) I just haven't had as much time as I generally like to poke around at it for a sweeping (if super tiny and completely reasonable) change.

taldcroft:
I recall there even being problems with the dark map being used in ASVT and requiring custom processing to fix the negative values.  Basically they are just wrong.  The only thing I could think of being a problem is some analysis script that is interested in the full distribution.  That could just be fixed on an as-discovered basis.

jeanconn:
And if we had more regression tests in proseco, the guide star selection that is doing spoiler rejection on the basis of "this spoiler star contributes more counts than the Nth percentile dark current" could show more changes, but star selection based on that would change the same whether we effect the change in proseco or mica.

taldcroft:
True but the only thing we would learn is that selection process to was impacted by the bug this now fixed.

jeanconn:
Agreed.

PR #190, opened 2019-02-17
Drop version file

taldcroft:
The whole fancy `version.py` thing has not proved useful and has been overtaken by `test(get_version=True)`.  This drops that file and bumps the version to 3.16 for the next release.


taldcroft:
Note that merging this doesn't require that we make a new release, but I'd kind of like to.  We'll see.

jeanconn:
Note that I had assumed that tests passed with this PR but had not run them myself.  I see now they don't pass.

PR #191, opened 2019-02-19
Fix use of mica.version and tests

jeanconn:
Fix module use of (removed) mica.version and fix up tests


jeanconn:
The missing mica.version fixes have not been well-tested, but at least passes test.

jeanconn:
Great.  I think mica.version also needs to be removed from setup.py and conf.py (just grepping around now too).

PR #197, opened 2019-06-26
Add support for acdc dark cal files

taldcroft:
As noted at https://github.com/sot/acdc/pull/16#issuecomment-505983796, the acdc JSON properties file doesn't play with the current mica.  This should fix that.  In the mixed case, one gets back a masked table with the combined set of scalar attributes from the two different formats.

### DIFF
--- a/mica/archive/tests/test_aca_dark_cal.py
+++ b/mica/archive/tests/test_aca_dark_cal.py
@@ -78,3 +78,29 @@ def test_get_dark_cal_props(aca_image, allow_negative):
 def test_get_dark_cal_props_table():
     props = dark_cal.get_dark_cal_props_table('2007:001', '2008:001')
     assert np.allclose(props['eb'], [24.6, 25.89, 51.13, 1.9])
+    assert props.colnames == ['ccd_temp', 'date', 'dec', 'dur', 'eb', 'el', 'id', 'l_l0',
+                              'ra', 'start', 'stop', 'sun_el', 'zodib']
+
+
+@pytest.mark.skipif('not HAS_DARK_ARCHIVE', reason='Test requires dark archive')
+def test_get_dark_cal_props_table_acdc():
+    """Just acdc dark cals, giving a non-masked result
+    """
+    props = dark_cal.get_dark_cal_props_table('2019:150', '2019:160')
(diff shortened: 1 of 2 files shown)


jeanconn:
@mbaski do you use the mica.archive.aca_dark.get_dark_cal_props_table method at all?  My guess is a strong no.

jeanconn:
On this side, it looks like this was used in the aca_dark_cal project for some fetching of data to make the reports.  That's all deprecated now so that's fine.  There are probably some other uses in a notebook here or there which we can deal with as needed.

jeanconn:
I'm not really sure if making the combined table will be useful for anything, but at least calling the method works for me now to not throw an error.  Thanks for the fix @taldcroft .  I think this would be fine to install this week.

mbaski:
> 
> 
> @mbaski do you use the mica.archive.aca_dark.get_dark_cal_props_table method at all? My guess is a strong no.

No (I searched my files and it did not come up and I don't recall ever using it).

mica/archive/tests/test_aca_dark_cal.py
```diff
@@ -78,3 +78,29 @@ def test_get_dark_cal_props(aca_image, allow_negative):
...
     assert np.allclose(props['eb'], [24.6, 25.89, 51.13, 1.9])
+    assert props.colnames == ['ccd_temp', 'date', 'dec', 'dur', 'eb', 'el', 'id', 'l_l0',
+                              'ra', 'start', 'stop', 'sun_el', 'zodib']
+
+
+@pytest.mark.skipif('not HAS_DARK_ARCHIVE', reason='Test requires dark archive')
+def test_get_dark_cal_props_table_acdc():
+    """Just acdc dark cals, giving a non-masked result
+    """
+    props = dark_cal.get_dark_cal_props_table('2019:150', '2019:160')
```
jeanconn: I note I missed that I think these new tests do not work with the current archive until we promote the 2019:158 dark cal into it.  I think I'm OK with that.

taldcroft: Yeah that was a bit of a chicken and egg that I could not figure a way around.

PR #198, opened 2019-06-27
Remove update_aca_dark pieces

jeanconn:
Remove update_aca_dark.py and related pieces


jeanconn:
I was just going to edit the cron task, but don't see a reason to keep the script either.

jeanconn:
I could use a thumbs up or down on this now so I can cut 3.18 mica and make stuff to brief it for LR.

jeanconn:
Thanks!

jeanconn:
Though now I'm debating if I should install 3.18 on ska2 flight or just manually cut the cron task.

taldcroft:
The truth is that you don't *need* to do any of this from my understanding.  As long as you do not manually install a new dark cal into /proj/sot/ska/data/aca_dark_cal then the mica code will do nothing.

jeanconn:
Well, removing the entry in the task schedule in version control is a good idea (and is in here too) but the removing the rest of the module pieces is not required as you note.  Regarding 3.18 it was really a question on if we wanted the updated get_dark_cal_props_table to work in ska2 .

PR #200, opened 2019-06-27
Skip tests that really weren't made for PY2

jeanconn:
Skip tests that really weren't made for PY2

I suppose this includes one test that used to work and now doesn't but we don't care.  If I install on ska2 I'd just like to not see fails and this seemed the least amount of work to accomplish that.

### DIFF
--- a/mica/archive/tests/test_aca_dark_cal.py
+++ b/mica/archive/tests/test_aca_dark_cal.py
@@ -5,11 +5,13 @@ Basic functionality and regression tests for ACA dark cal module.
 import os
 import numpy as np
 import pytest
+import six
 
 from ..aca_dark import dark_cal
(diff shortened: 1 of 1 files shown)


jeanconn:
This is all for the get_dark_cal_props_table issue as well.

taldcroft:
Why are we spending time on this?

jeanconn:
For the same reason that I just emailed.  I'd like ska_testr tests to pass for modules that are still being used for cron jobs. 

That way if you do update a data product that is still being used by a job/module we still actually use in ska2, I can run our regression tests  and a fail will actually be something I'll care about, instead of then trying to figure out what has changed and if it is really important or ...

taldcroft:
Can we close this?

PR #201, opened 2019-06-27
Initial infrastructure for mica cron processing in Ska3

taldcroft:
These are untested changes suggesting a path toward doing mica cron processing in Ska3 and keeping the Ska2 mica entirely static going forward.  It takes lessons from `acis_taco` in terms of package and data architecture.  The basic idea is:

- Stop any chance of accidental mica dark cal updates by simply commenting out that task in the live `/proj/sot/ska/share/mica/task_schedule.cfg`.  There won't be updates anyway if no new dark cals are added to `$SKA/data/aca_dark_cal`, but it does not hurt to be safe.
- Stop Ska2 processing of starcheck updates by commenting out that task in the same way.
- Start processing of starcheck updates in Ska3 with this PR.

This PR also includes moving starcheck parsing to starcheck where it makes more sense.  This change is not really required for the Ska3 transition, but as long as big things are happening it doesn't hurt to toss this in.  Requires https://github.com/sot/starcheck/pull/326.

This is all painful and way more work than the alternative of just fixing ska2 for now, but it has to be done at some point so it is not wasted effort and it seems like now is fine time to start.


jeanconn:
For this, the mica.starcheck changes fine.  I don't see what the database initialization changes really buy us in the other pieces at this point.

taldcroft:
The DB init changes are needed to remove the makefile and make the conda package complete.  At least that is the intent.

jeanconn:
I just don't have much to test / verify those changes so was wondering if they should just be packaged with individual tool updates (update aca_l0 processing, update aca_l0 processing etc).  This is fine as long as I remember to reference these changes if something breaks.

task_schedule.cfg
```diff
@@ -39,17 +39,20 @@ alert       aca@head.cfa.harvard.edu
...
 <task mica_archive>
       cron       * * * * *
       check_cron * * * * *
-      exec 1: update_obspar.py
-      exec 1: update_aca_l0.py
-      exec 1: update_asp_l1.py
+      # exec 1: update_obspar.py
+      # exec 1: update_aca_l0.py
+      # exec 1: update_asp_l1.py
       exec 1: update_starchecks.py
```
jeanconn: So my plan was to put this in its own task schedule file ... would you prefer I do it that way or set up multiple tasks in a single task schedule (we haven't really done that in a while)?

taldcroft:
> I just don't have much to test / verify those changes so was wondering if they should just be packaged with individual tool updates (update aca_l0 processing, update aca_l0 processing etc). This is fine as long as I remember to reference these changes if something breaks.

It's just more difficult to remember to do this in the same way for each tool and makes overall LOE larger.  Since these tools are basically not used anymore, and definitely not in production processing, all you need to do now for testing/verification is to confirm that the sql init files are installed in the expected place in the package (along-side the respective modules).   By inspection the code will then work.  If not, we fix.

task_schedule.cfg
```diff
@@ -11,7 +11,7 @@ iterations 1
 
 data_dir     $ENV{SKA_DATA}/mica/      # Data file directory
 log_dir      $ENV{SKA_DATA}/mica/logs  # Log file directory
-bin_dir      $ENV{SKA_SHARE}/mica/     # Bin dir (optional, see task def'n)
+bin_dir      $ENV{SKA_ARCH_OS}/share/mica      # Bin dir (optional, see task def'n)
```
jeanconn: I thought we decided SKA_ARCH_OS was deprecated so could tweak this https://github.com/sot/skare3/issues/22#issuecomment-412646934

jeanconn:
OK.  With regard to testing, this isn't really setup to do incremental tests, so I didn't bother.  A full rebuild of the starcheck database using this code left me with a starcheck_catalog table that had all the same rows as the flight table.  I think we can remove the WIP and just go with this.

jeanconn:
I do get errors if I try to pip install in my dev environment with
```
 pip install --no-deps --verbose --no-binary :all: --no-index  .
...Command "/fido.real/ska3/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-bo8s30zy-build/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-apqwzhda-record/install-record.txt --single-version-externally-managed --compile" failed with error code 1 in /tmp/pip-bo8s30zy-build/
Exception information:
Traceback (most recent call last):
  File "/fido.real/ska3/lib/python3.6/site-packages/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/fido.real/ska3/lib/python3.6/site-packages/pip/commands/install.py", line 342, in run
    prefix=options.prefix_path,
  File "/fido.real/ska3/lib/python3.6/site-packages/pip/req/req_set.py", line 784, in install
    **kwargs
  File "/fido.real/ska3/lib/python3.6/site-packages/pip/req/req_install.py", line 878, in install
    spinner=spinner,
  File "/fido.real/ska3/lib/python3.6/site-packages/pip/utils/__init__.py", line 707, in call_subprocess
    % (command_desc, proc.returncode, cwd))
pip.exceptions.InstallationError: Command "/fido.real/ska3/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-bo8s30zy-build/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-apqwzhda-record/install-record.txt --single-version-externally-managed --compile" failed with error code 1 in /tmp/pip-bo8s30zy-build/
```

task_schedule.cfg
taldcroft: Indeed that is correct.  The only hesitation I find now (with experience a year later) is that using `$SKA` here is potentially confusing.  Yes, since this will be running in a production environment so `$SKA` will always work, but now that I mostly run in a standalone environment that *looks* confusing because I immediately think that `$SKA` is outside the installation environment where `share/mica` actually lives.  So maybe sticking with `$SKA_ARCH_OS` is not such a bad thing in this context because it gives an clear and distinct name to what we mean.

jeanconn: I don't think it is very distinct?  Fine with this PR as-is but still confused this then.

taldcroft: `$SKA_ARCH_OS` is distinct because it *only* exists a production environment and can only mean the root of the runtime environment directory.  This more clearly signals what's happening.  `$SKA` on the other hand is more ambiguous.  The Ska3 "always-true" definition of `$SKA` is where `$SKA/data` lives, but in a production env that happens to coincide with the root of the runtime env.

jeanconn: We should probably bring this back to the SKA thread.  Why does SKA_ARCH_OS only exist in a production environment?  I didn't think flt_envs required that but it sounds like you have that convention in mind?

taldcroft:
With  68db5d1 it looks like the scripts and sql init files are all in the right place.
```
(ska3-mica) neptune$ pip install --no-deps --no-binary :all: --upgrade --no-index  .
Processing /Users/aldcroft/git/mica
Installing collected packages: mica
  Found existing installation: mica 4.18
    Uninstalling mica-4.18:
      Successfully uninstalled mica-4.18
  Running setup.py install for mica ... done
Successfully installed mica-4.18

(ska3-mica) neptune$ ls /Users/aldcroft/miniconda3/envs/ska3-mica/share/mica/
task_schedule.cfg      update_acq_stats.py*   update_guide_stats.py* update_reports.py*     update_vv.py*
update_aca_l0.py*      update_asp_l1.py*      update_obspar.py*      update_starchecks.py*

(ska3-mica) neptune$ ls /Users/aldcroft/miniconda3/envs/ska3-mica/lib/python3.6/site-packages/mica/archive/
__init__.py                aca_l0.py                  asp_l1_bad_obsids.dat      obspar.py
__pycache__/               archfiles_aca_l0_def.sql   asp_l1_proc.py             obspar_def.sql
aca_dark/                  archfiles_asp_l1_def.sql   cda_aca0_list.py           processing_asp_l1_def.sql
aca_hdr3.py                asp_l1.py                  obsid_archive.py           tests/
```

task_schedule.cfg
taldcroft: Sorry, I've been sloppy with terminology, I was equating "production environment" with "sourcing flt_envs on HEAD linux", or maybe "not standalone".  Do you understand my statement now?

jeanconn: Still not quite there.  SKA_ARCH_OS is defined in flt_envs.  Do you explicitly not use flt_envs in most of your work at this point and just set SKA manually, even though flt_envs theoretically could be run for any ska3 conda environment?  Should we move to equating use of flt_envs with a "production" environment?  

I see your point that it doesn't much matter because SKA_ARCH_OS should be defined in HEAD ska3/flight, I just didn't see that as quite distinct from any other ska environment just yet, but maybe that's just how I'm using things.

taldcroft: I always set `export SKA=${HOME}/ska` and that's it.  That's where my Ska data live.  If you run `flt_envs` on a standalone environment that basically breaks Ska3 since `$SKA` is no longer pointing at your data (unless you do some shenanigans with a link).

taldcroft: > We should probably bring this back to the SKA thread. 

That's done (in fact I did that before you suggested it!).

jeanconn: Where?  I was wondering about updating the text in https://github.com/sot/skare3/issues/22

taldcroft: https://github.com/sot/skare3/issues/22#issuecomment-412646934

But now I see the topic description needs tweaking to reflect the new approach on SKA_ARCH_OS.

PR #206, opened 2019-08-28
Add update_agasc_supplement.py script

taldcroft:
This provides a configured way to manually update the `agasc_supplement.h5` file to include a new bad star.  The prototype of 839258056 from obsid 22728 is the driver.

See also the companion wiki page: https://github.com/sot/mica/wiki/AGASC-supplement

This is well short of a nice automated system, but hopefully will suffice for now.

### DIFF
--- a/scripts/update_agasc_supplement.py
+++ b/scripts/update_agasc_supplement.py
@@ -1,0 +1,99 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+"""
+Update the agasc_supplement.h5.
+
+This file is a supplement to the stable AGASC to inform star selection
+and star catalog checking.
+
+Currently this script only has the capability to add a bad star to the
+bad star table.  It might end up including functionality to automatically
+update another table with effective mags based on acq / guide history.
+"""
+import os
+import argparse
+from pathlib import Path
+
+import pyyaks.logger
+from astropy.table import Table
+
+SKA = Path(os.environ['SKA'])
+logger = None  # Set via global in main()
+
+
+def get_options(args=None):
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--data-root",
+                        default='.',
+                        help=("Directory containing agasc_supplement.h5 (default='.')"))
+    parser.add_argument("--bad-star-id",
+                        type=int,
+                        help="AGASC ID of star to add to bad-star list")
+    parser.add_argument("--bad-star-source",
+                        type=int,
+                        help=("Source identifier indicating provenance (default=max "
+                              "existing source + 1)"))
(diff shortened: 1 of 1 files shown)


jeanconn:
OK.  So the plan is to just increment a ids to a "bad source" as we go to reference why a star has been added to the bad list?  Seems fine.

I had been wondering for 22728 if we wanted to add it to the bad list or try to get some magnitude estimate to use in a mag supplement table in this file (it looks to me like it is a 10.5 mag star).  Looks like we aren't wanting for ACQ stars in the field so fine to exclude it for both acq and guide then.

scripts/update_agasc_supplement.py
```diff
@@ -0,0 +1,99 @@
...
+    # Set up logging
+    loglevel = int(opt.log_level)
+    logger = pyyaks.logger.get_logger(name='mica_update_agasc_supplement', level=loglevel,
+                                      format="%(message)s")
+
+    # If --data-root is supplied then set the fetch msid_files basedir via ENG_ARCHIVE
+    # prior to importing fetch.  This ensures that ``content`` is consistent with
+    # the destination archive.
+    if opt.data_root is None:
+        data_root = SKA / 'data' / 'agasc'
```
jeanconn: This is a case where I'm actually not thrilled with it defaulting to SKA even if you've warned "don't have SKA set to /proj/sot/ska when you do this or don't be on HEAD SKA".  I would prefer that the directions and defaults  and maybe process were fine if working as the aca user in flight ska3.

taldcroft: Note that the default for `opt.data_root` is now `'.'` now, so this branch of the logic cannot be reached (I think).  So there is some code cleanup and corresponding update to instructions needed.

But more to the point, the driver for the weird instructions is that I cannot see how to fully test this file (within proseco) without having it "installed" in `$SKA/data/agasc` because proseco is hardwired to look for it there.  So working as aca in flight ska3 is not a good option.

jeanconn: Gotcha.  Thanks!   And no big deal then, but maybe it would make just as much sense for the script to take an explicit input file and write an explicit output file (and then let user make or use appropriate test environment).

taldcroft: I updated the script and process document for a flow that I think is reasonably simple.  I don't think you need to specify the entire file name since it really needs to be `agasc_supplement.h5` to be useful.

jeanconn: OK.  I'd still prefer it took old file as input and made a new file as output but if you need the script to update the file in-place this is fine.

taldcroft: Noted.  You can always make a copy in advance of the in-place update to get the same effect.

PR #212, opened 2019-12-30
Fix an issue in PCAD table viewer with django 2+

taldcroft:
This import style (instead of "from kadi import events") is required due to a conflict with how the django app is initialized and standalone usage since django 2+. Basically `events/__init__.py` in a django app does not get the auto-generated query references because DJANGO_SETTINGS_MODULE is set.

## Testing

This works in the ska3-django3 test environment using the django dev server:
- [x] Linux
- [x] MacOS


taldcroft:
This change is so small and uncontroversial that I am merging without review.

PR #214, opened 2020-02-03
Remove dark model

taldcroft:
## Description

Remove the `dark_model` module which has been deprecated for a while.  There are no instances of dark_model coming from mica in the `sot` org.

This is causing warnings now in pytest every time `mica.archive.aca_dark` gets imported.

## API impact

Any code now importing dark modeling functions from `mica.archive.aca_dark` will now break.

## Testing

- [x] Passes unit tests on MacOS(with a fair number of skips)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,4 @@
 doc/_build
 *.pyc
 *~
-
+*.egg-info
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -3,4 +3,4 @@
 doc/_build
 *.pyc
 *~
-
+*.egg-info
```
jeanconn: I rebased this to clear the .gitignore conflict.  Still seems fine to go.

PR #218, opened 2020-03-24
Minor PY3 updates and DS10.8.3 update for mica vv processing

### LINKED ISSUE #202: Support Ska3 cron processing
Finish the job started by #201.

jeanconn:
## Description

Minor PY3 updates and DS10.8.3 update for mica vv processing

## Testing

- [x] Functional testing (which was running a few of these pieces in an "update" mode in a test MICA_ARCHIVE

Fixes #202


mica/archive/obsid_archive.py
```diff
@@ -39,9 +39,10 @@ def parse_obspar(file):
                'r': float,
                's': str}
     try:
-        lines = gzip.open(file).readlines()
+        lines = gzip.open(file, 'rb').readlines()
```
taldcroft, line 42: You should be able to put the `encoding` and `errors` in the calls to `gzip.open` and `open`.

mica/report/report.py
```diff
@@ -319,9 +320,8 @@ def obs_links(obsid, sequence=None, plan=None):
     # if this is a science observation, only try to get a star catalog if it has a home
     # in the schedule either in the past or the near future
     if plan is not None:
-        plan_date = DateTime("{:4d}:{:03d}:{:02d}:{:02d}:{:02d}.000".format(
-                plan.year, plan.dayofyear, plan.hour, plan.minute, plan.second))
-        if plan_date.secs < (DateTime() + 21).secs:
+        plan_date = Time(plan)
+        if plan_date.cxcsec < (DateTime() + 21).secs:
```
taldcroft, line 324: The astropy-way to write this would be:
```
if plan_date < Time.now() + 21 * u.day:
    ...
```

mica/report/report.py
```diff
@@ -512,7 +512,7 @@ def main(obsid):
         if summary is None:
             raise ValueError("Obsid not found in target table")
         report_status['ocat'] = summary['status']
-        links = obs_links(obsid, summary['seq_nbr'], summary['lts_lt_plan'])
+        links = obs_links(obsid, summary['seq_nbr'], str(summary['lts_lt_plan']))
```
taldcroft, line 515: Unicode sandwich?  Is Ska.DBI.DBI returning `bytes` for a Sybase query?  I'm a little surprised, but if so we should fix that upstream in `Ska.DBI`.

mica/vv/core.py
```diff
@@ -284,7 +286,7 @@ def _save_info_json(self, file=None):
     def _save_info_pkl(self, file=None):
         if file is None:
             file = os.path.join(self.tempdir, 'vv_report.pkl')
-        pfile = open(file, 'w')
+        pfile = open(file, 'wb')
```
taldcroft, line 289: Context manager.

mica/report/report.py
```diff
@@ -117,7 +118,7 @@ def rec_to_dict_list(recarray):
 
 
 def get_star_acq_stats(id):
-    hdu = tables.open_file(acq_stats.table_file)
+    hdu = tables.open_file(acq_stats.TABLE_FILE)
```
taldcroft, line 121: What's happening with this change?

mica/archive/asp_l1_proc.py
```diff
@@ -53,8 +53,12 @@ def update(obsids, config=None):
         for sol in asols:
             logger.info("\tprocessing {}".format(sol))
             procdir = os.path.dirname(sol)
-            logfile = glob(os.path.join(procdir, "*log*"))[0]
-            aspect_log = gzip.open(logfile).read()
+
+            # As of DS 10.8.3, there are both "com" logs and per-ai logs.
+            # This glob should get the per-ai logs.
+            logfile = glob(os.path.join(procdir, "asp_l1_f*log*"))[0]
```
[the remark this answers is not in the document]
jeanconn: Why do I need to check the number of matchin files?

mica/archive/obsid_archive.py
jeanconn: Sure.  We had just had those comments about what is in scope.  This also applies to the use of context managers.

mica/report/report.py
jeanconn: Not bytes.  Ska.DBI is returning a datetime.datetime (I think) and the downstream handling is a little different.

mica/report/report.py
jeanconn: I don't remember when the name changed in acq stats.

mica/archive/aca_l0.py
```diff
@@ -460,12 +460,14 @@ def __init__(self,
 
     def _get_missing_archive_files(self, start, only_new=False):
         ingested_files = self._get_arc_ingested_files()
+        ingest_dates = [file['ingest_date'].decode() for file in
```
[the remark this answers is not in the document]
jeanconn: OK.  I'll try again.  I think I *did* try Tables to start but ened up thinking that the decode() was actually more minimal.

mica/archive/asp_l1_proc.py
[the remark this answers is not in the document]
jeanconn: The number of files I get that matches the pattern is the right number.

mica/report/report.py
jeanconn: In retrospect, it is probably https://github.com/sot/mica/pull/173/files but still just need to get the reference to match whatever it is by the time we promote this code.

mica/archive/asp_l1_proc.py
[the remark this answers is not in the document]
taldcroft: Your code is *assuming* that one and only one file matches.  Better to use an existing well-defined interface that provides proper checking and useful exception messages.

mica/archive/obsid_archive.py
taldcroft: I think of this as working on an old house, so any wiring or plumbing that gets touched should be brought up to code.  As long as there are no fires burning we can take the time to do things right, reinforcing good practices and making incremental improvements.

mica/archive/asp_l1_proc.py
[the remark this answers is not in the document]
jeanconn: Oh, you mean with the `[0]` bit that has not changed in this PR.

mica/archive/obsid_archive.py
jeanconn: Sure.  I just wasn't sure about level of fires burning.  The ingest is currently broken which actually makes it harder to do things like V&V, and I wasn't sure if I'd catch load review this week.  I'm feeling slow.

Also, if we need LR approval for all ska3 code, we should have broken mica up into a separate project for each of these tasks because I still figure I'm going to bring this for review and then notice a tiny issue when run in flight that will keep everything from working.

mica/report/report.py
taldcroft: Astropy.Time accepts `datetime` , so maybe with your new code it just works.  Bigger point though is to include comments on any code that is not obvious, like why is something being cast as `str` (esp. in this context when it used to work without the cast).

mica/archive/obsid_archive.py
taldcroft: We can get this (and subsequent bug fixes) quickly approved out-of-cycle if need be.

mica/report/report.py
jeanconn: I suppose I should just comment up the PR more for that?  It seems odd to have a comment in the code like "the python 2 interface to sql was returning something different here but now that it is a datetime it seems less trouble to just stringify it when passing to a routine that previously was just getting a string".

mica/archive/asp_l1_proc.py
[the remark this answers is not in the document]
taldcroft: Same "fixing the old house" comment.  Easy incremental improvements should just be done.  Esp. in this case when upstream (DS) processing has changed in a substantive way with different log files, maybe we will be surprised.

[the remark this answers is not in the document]
jeanconn: Yes, I'll make improvements.  Here, however, the code is I think only reading a per-ai log file just to try to get some kind of version.  If DS surprised us with multiple files I would hope they would still have one version.  So I'm trying to think of actionable items when adding more checking.

mica/report/report.py
taldcroft: An appropriate comment might be
```
# Note that summary['lts_lt_plan'] is a datetime but `obs_links()` requires a str
```
From the original code I'd say that this `lts_lt_plan` was always a `datetime` (from the attributes that get used).  I'm not even sure why it would fail now without the `str()` wrapper.

jeanconn: Yeah I was planning to go with your "maybe it just works" and shouldn't have fought over the WIP implementation when I was probably going to come back these.  The only real issue with datetime I seem to have hit was that the py3 datetime.datetime was missing dayofyear .

taldcroft: FYI it's a little obscure, but day of year is there as yday in the `timetuple()` output: https://docs.python.org/3/library/datetime.html#datetime.date.timetuple.

jeanconn: Thanks!  But the code was using dayofyear to build YYYY:DOY:HH:MM:SS.S , and I think that is now unnecessary.

mica/archive/aca_l0.py
```diff
@@ -35,11 +36,11 @@
                       'tlmver', 'ascdsver', 'revision', 'date',
                       'imgsize')
 
-
-FDTYPE = [('level', '|S4'), ('instrum', '|S7'), ('content', '|S13'),
-         ('arc5gl_query', '|S27'), ('fileglob', '|S9')]
-FILETYPE = np.rec.fromrecords([('L0', 'PCAD', 'ACADATA', 'ACA0', '*fits.gz')],
-                              dtype=FDTYPE)[0]
+FILETYPE = {'level': 'L0',
```
jeanconn, line 39: The np.rec.fromrecords gave us unnecessary bytestrings.  I *think* this needs no new comments as the "this used to be bytestrings" part isn't interesting and the simple dictionary to store the type seems self-documenting.

mica/archive/aca_l0.py
```diff
@@ -465,7 +466,7 @@ def _get_missing_archive_files(self, start, only_new=False):
                     startdate)
         # find the index in the cda archive list that matches
         # the first entry with the "start" date
-        for idate, backcnt in zip(ingested_files[::-1]['ingest_date'],
+        for idate, backcnt in zip(ingested_files['ingest_date'][::-1],
```
jeanconn, line 469: Looks like I just wasn't paying attention to order when I put this "back" after doing more of the "unicode sandwich" but this order is probably better anyway.

PR #219, opened 2020-03-27
Mag stats

### LINKED ISSUE #216: Add guide star 5th, 16th, 50th, 84th, and 95th percentile mag columns to guide stats.
Let's just do this now and start reprocessing the mission for https://github.com/sot/ska-projects/issues/114.

jeanconn:
## Description

Add 5, 16, 50, 84, 95 mag stats to the guide stats table.  This also adds some changes to support running on time ranges to make it easier to do trivial multiprocessing.

## Testing

- [x] Functional testing

Fixes #216

### DIFF
--- a/mica/stats/update_guide_stats.py
+++ b/mica/stats/update_guide_stats.py
@@ -112,7 +112,7 @@ GUIDE_COLS = {
         ('var', 'int'),
         ('pos_err', 'int'),
         ('mag_aca', 'float'),
-        ('mag_err', 'int'),
+        ('mag_aca_err', 'int'),
         ('mag_band', 'int'),
         ('pos_catid', 'int'),
         ('aspq1', 'int'),
--- a/mica/stats/update_guide_stats.py
+++ b/mica/stats/update_guide_stats.py
@@ -135,12 +135,6 @@ GUIDE_COLS = {
     }
 
 
-SKA = os.environ['SKA']
-DATA = os.path.join(SKA, 'data', 'guide_stats')
-TABLE_FILE = os.path.join(DATA, 'guide_stats.h5')
-SKIPPED_FILE = os.path.join(DATA, 'skipped.dat')
-
-
 def get_options():
     parser = argparse.ArgumentParser(
         description="Update guide stats table")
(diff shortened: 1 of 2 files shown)


mica/stats/update_guide_stats.py
```diff
@@ -451,7 +454,14 @@ def table_gui_stats(obsid_info, gui_stats, star_info, catalog, temp):
     logger.info("arranging stats into tabular data")
     cols = (GUIDE_COLS['obs'] + GUIDE_COLS['cat'] + GUIDE_COLS['stat']
             + GUIDE_COLS['agasc'] + GUIDE_COLS['temp'] + GUIDE_COLS['bad'])
+
+    # Initialize all values to zero
     table = Table(np.zeros((1, 8), dtype=cols).flatten())
+    # Set all columns with mag info to 99.0 initial value instead of zero
+    for col in table.dtype.names:
+        if 'aoacmag' in col:
+            table[col] = 99.0
```
jeanconn: I think this is somewhat reasonable to "initialize" the numpy table with values of 99 for all the aoacmag_* cols if we want 99.0 for that kind of missing data.  Not sure if other columns should get that treatment too.

jeanconn: This is probably the only controversial thing in here.  As with everything, hard to know how much to rummage around and improve things and how much to just back away slowly.

mica/stats/update_guide_stats.py
```diff
@@ -112,7 +115,7 @@
         ('var', 'int'),
         ('pos_err', 'int'),
         ('mag_aca', 'float'),
-        ('mag_err', 'int'),
+        ('mag_aca_err', 'int'),
```
jeanconn, line 118: Hi @taldcroft  and @javierggt ! It sounds like you also want a change in this PR somewhere to scale the mag_aca_err column to units of mag.  That's fine.  Though should that be in mica or in agasc.get_star (or get_stars) with a calibrate flag or some such?

taldcroft: Well given that all the rest of this is uncalibrated it appears we cannot make a change here without having a one-off inconsistency.  Just leave it.  And don't bother with making an `agasc` issue.  If one of us gets annoyed enough one day we'll do it, but it is a low priority and we don't need more low priority issues.

jeanconn: Would it be less inconsistent to just UPPER CASE these?

taldcroft: It would be more consistent to upper case them, but would break all existing code that uses these.  Not worth the trouble I'd say.

PR #221, opened 2020-05-09
Recast dark cal image to native byte order

taldcroft:
## Description

This implements the right upstream fix for https://github.com/sot/chandra_aca/issues/97.

## Testing

- [x] Passes unit tests on MacOS (both mica and chandra_aca using this version of mica)
- [x] Functional testing:
```
In [1]: from mica.archive import aca_dark
In [2]: ccd_bgd = aca_dark.dark_cal.get_dark_cal_image('2019:001', aca_image=True)
In [3]: ccd_bgd.flicker_update(1)
In [4]:  # No failure
```

Fixes https://github.com/sot/chandra_aca/issues/97

### DIFF
--- a/mica/archive/aca_dark/dark_cal.py
+++ b/mica/archive/aca_dark/dark_cal.py
@@ -122,9 +122,11 @@ def _get_dark_cal_image_props(date, select='before', t_ccd_ref=None, aca_image=F
     """
     DARK_CAL['id'] = get_dark_cal_id(date, select)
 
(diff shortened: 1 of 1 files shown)


jeanconn:
Great idea!  Though should we integration test with Windows proseco before merging?

taldcroft:
This is very unlikely to behave differently on Windows or break anything, and if it does end up producing a regression we'll see it in the next MATLAB release testing and fix it then.

jeanconn:
Agreed.  I just wasn't sure how easy it would be to trace back to this change if it did something weird.

PR #222, opened 2020-05-19
Change sybase skip test to check for more restrictive vv creds

jeanconn:
Change sybase skip test to check for more restrictive vv creds

This should let the report-writing test that uses the axafvv query
run as jeanconn user, but skip for other users.

## Testing

- [x] Functional testing (the functional testing was the unit testing.  I ran as jeanconn, aca, and kadi users from master and from this branch and see expected test skips).

Fixes #

### DIFF
--- a/mica/report/tests/test_write_report.py
+++ b/mica/report/tests/test_write_report.py
@@ -8,7 +8,7 @@ from .. import report
 
 try:
     import Ska.DBI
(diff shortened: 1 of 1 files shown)


javierggt:
I'm ok with this, because I see checked it will be skipped in my case.

PR #225, opened 2020-06-23
Remove tables3_api imports

jeanconn:
## Description

Remove tables3_api imports.  Since we're no longer building tables3_api for shiny and it really isn't needed anymore (we're not doing Py2 backports) it is fine to just cut it here.

## Testing

I think the unit testing is sufficient.

- [ ] Functional testing 

Fixes #


taldcroft:
Ooops, didn't see this and independently made the changes in #226 .  I'll revert those.

Issue #179, opened 2018-07-26
Test failure running as not jeanconn


taldcroft:
Presumably I could have the right credentials, but this should probably just be skipped if the credentials file isn't there.
```
=================================== FAILURES ===================================
______________________________ test_write_reports ______________________________

    @pytest.mark.skipif('not HAS_SYBASE_ACCESS', reason='Report test requires Sybase/OCAT access')
    @pytest.mark.skipif('not HAS_SC_ARCHIVE', reason='Report test requires mica starcheck archive')
    def test_write_reports():
        """
        Make a report and database
        """
        tempdir = tempfile.mkdtemp()
        # Get a temporary file, but then delete it, because report.py will only
        # make a new table if the supplied file doesn't exist
        fh, fn = tempfile.mkstemp(dir=tempdir, suffix='.db3')
        os.unlink(fn)
        report.REPORT_ROOT = tempdir
        report.REPORT_SERVER = fn
        for obsid in [20001, 15175, 54778]:
>           report.main(obsid)

../../../../x86_64-linux_CentOS-6/lib/python2.7/site-packages/mica/report/tests/test_write_report.py:35: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../../x86_64-linux_CentOS-6/lib/python2.7/site-packages/mica/report/report.py:501: in main
    links = obs_links(obsid, summary['seq_nbr'], summary['lts_lt_plan'])
../../../../x86_64-linux_CentOS-6/lib/python2.7/site-packages/mica/report/report.py:342: in obs_links
    vv, vvid = official_vv(obsid)
../../../../x86_64-linux_CentOS-6/lib/python2.7/site-packages/mica/report/report.py:262: in official_vv
    vv_db = Ska.DBI.DBI(dbi='sybase', server='sqlsao', user='jeanconn', database='axafvv')
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <Ska.DBI.DBI.DBI object at 0x7ff9138e8d10>, dbi = 'sybase'
server = 'sqlsao', user = 'jeanconn', passwd = None, database = 'axafvv'
numpy = True, autocommit = True, verbose = False
[...]

Issue #152, opened 2017-07-20
V&V queries run as hardcoded 'jeanconn' user


jeanconn:
This should at least be set so 'taldcroft' can run it as well.

jeanconn:
This was fixed.  Tests still not skipping correctly for non-priv user but should work for user with right privs.

PR #231, opened 2020-08-07
More shiny fixes

### LINKED ISSUE #228: Failing unit test test_run_vv_omitted_fid in shiny on HEAD
Some test issues in shiny (seen with `pytest mica` from shiny branch)

- Failing unit test test_run_vv_omitted_fid in shiny on HEAD. This is passing in ska3-flight.
- DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead

jeanconn:
## Description

Fix V&V handling of missing fid in shiny (also works in ska3-flight). 
Also set code to assume USER to run the axafvv queries so the V&V report making can work as user other than 'jeanconn'.
Set report writing test to work as USER as well and provide new warnings on Linux if the axafvv database access doesn't work.

## Testing

- [x] Passes unit tests on Linux with ska3-flight or shiny 
- [x] Functional testing does not apply as the code changes were to remove warnings and errors in unit tests outputs

Fixes #228, #179, #152

### DIFF
--- a/mica/report/report.py
+++ b/mica/report/report.py
@@ -259,7 +259,8 @@ def guess_fot_summary(mp_dir):
 
 
 def official_vv(obsid):
-    vv_db = Ska.DBI.DBI(dbi='sybase', server='sqlsao', user='jeanconn', database='axafvv')
+    user = os.environ.get('USER') or os.environ.get('LOGNAME')
+    vv_db = Ska.DBI.DBI(dbi='sybase', server='sqlsao', user=user, database='axafvv')
     vv = vv_db.fetchone("""select vvid from vvreport where obsid = {obsid}
                            and creation_date = (
                               select max(creation_date) from vvreport where obsid = {obsid})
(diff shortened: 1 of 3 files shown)


taldcroft:
Does this pass tests as a user without sybase access?

jeanconn:
Good question.  I think the logic I put in for the sybase stuff is moving in the right direction, but I suppose I can try tests as kadi user on linux and see if I get the right skips and expected warns.

jeanconn:
So stringifying the fid id seems to be the only thing needed to pass tests.

The changes to remove the "jeanconn" business look to me like the work to add new warnings if the tests are being run on head either in an environment that isn't set up for sybase or as a user who doesn't have axafvv database access. Passes test as 'kadi' user with and without SYBASE set, appropriately skips the report test but adds a warning.  I *think* the warning is appropriate for my use cases but if we want to, say, pull the report-writing test into something separate to run as a "long" test we could do that.  We could also ask for 'aca' user to have 'axafvv' read access and Javier is going to need it too.

PR #233, opened 2020-08-13
Use q_att_raw for ASPSOL quaternions if available

taldcroft:
## Description

Make `get_atts` and friends work with modern ASPSOL data where the desired column is `q_att_raw` not `q_att`.

## Testing

- [ ] Passes unit tests on MacOS (only ran asp_l1 tests) [TESTS RUNNING at this point]
- [x] Functional testing

Via a temporary print statement, confirmed that `q_att_raw` is being used for obsid 23315 (recent) and `q_att` is being used for obsid 8008 (old).

### DIFF
--- a/mica/archive/asp_l1.py
+++ b/mica/archive/asp_l1.py
@@ -215,7 +215,8 @@ def get_atts_from_files(asol_files, acal_files, aqual_files, filter=True):
         # Quaternion multiply the asol quats with that inv misalign and save
         # I could also do this with the transform matrix and then only need
         # one accessory quat function.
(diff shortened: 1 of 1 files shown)


taldcroft:
I'm slightly confused about your point 1, just in the sense that the mica VV report for e.g. 23315 looks A-OK. Is that not what you are referring to? When I was searching around in the mica code I ran into that mapping that accounts for the various "raw" versions of columns, so I was under the impression it is all just working.

jeanconn:
My point was that the code to do residuals is duplicated.  That's not ideal and why I missed this for the centroid dashboard use case.

PR #241, opened 2020-09-23
Use vectorized Quat

jeanconn:
## Description

Use modern vectorized Quaternion.Quat instead of mica.quaterion

## Testing

Did not do additional function testing except for spot-checking a few of the intermediate matrices to see if they looked reasonable compared to master.

Fixes #

### DIFF
--- a/mica/archive/asp_l1.py
+++ b/mica/archive/asp_l1.py
@@ -13,7 +13,7 @@ import os
 import logging
 import numpy as np
 from astropy.table import Table
-from mica.quaternion import Quat
+from Quaternion import Quat
 
 from mica.archive import obsid_archive
 from mica.archive import asp_l1_proc
@@ -215,13 +215,13 @@ def get_atts_from_files(asol_files, acal_files, aqual_files, filter=True):
                        & (asol['time'] <= (aqual['time'][idx] + 1.025)))
                 asol = asol[~nok]
         # Make a Nx4 list of the inv misalign quats
-        q_mis_inv = np.repeat(Quat(acal['aca_misalign'][0]).inv().q,
+        q_mis_inv = np.repeat(Quat(transform=acal['aca_misalign'][0]).inv().q,
                               len(asol)).reshape((4, len(asol))).transpose()
         # Quaternion multiply the asol quats with that inv misalign and save
         # I could also do this with the transform matrix and then only need
         # one accessory quat function.
         q_att_name = 'q_att_raw' if 'q_att_raw' in asol.colnames else 'q_att'
-        att_chunks.append((Quat(asol[q_att_name]) * Quat(q_mis_inv)).q)
+        att_chunks.append((Quat(q=asol[q_att_name]) * Quat(q=q_mis_inv)).q)
         time_chunks.append(np.array(asol['time']))
         records.append(asol.meta)
     if len(att_chunks) > 0 and len(time_chunks) > 0:
--- a/mica/archive/asp_l1.py
+++ b/mica/archive/asp_l1.py
@@ -214,14 +214,17 @@ def get_atts_from_files(asol_files, acal_files, aqual_files, filter=True):
                 nok = ((asol['time'] >= (aqual['time'][idx] - 1.025))
                        & (asol['time'] <= (aqual['time'][idx] + 1.025)))
                 asol = asol[~nok]
-        # Make a Nx4 list of the inv misalign quats
-        q_mis_inv = np.repeat(Quat(transform=acal['aca_misalign'][0]).inv().q,
-                              len(asol)).reshape((4, len(asol))).transpose()
+
+        # Transpose of transform/rotation matrix is the inverse
+        aca_mis_inv = acal['aca_misalign'][0].transpose()
+
+        # Repeat the transform to match asol
+        aca_mis_inv = np.repeat(aca_mis_inv[np.newaxis, ...], repeats=len(asol), axis=0)
+        q_mis_inv = Quat(transform=aca_mis_inv)
+
         # Quaternion multiply the asol quats with that inv misalign and save
-        # I could also do this with the transform matrix and then only need
-        # one accessory quat function.
         q_att_name = 'q_att_raw' if 'q_att_raw' in asol.colnames else 'q_att'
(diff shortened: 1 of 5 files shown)


mica/archive/asp_l1.py
```diff
@@ -210,13 +210,13 @@ def get_atts_from_files(asol_files, acal_files, aqual_files, filter=True):
                        & (asol['time'] <= (aqual['time'][idx] + 1.025)))
                 asol = asol[~nok]
         # Make a Nx4 list of the inv misalign quats
-        q_mis_inv = np.repeat(Quat(acal['aca_misalign'][0]).inv().q,
+        q_mis_inv = np.repeat(Quat(transform=acal['aca_misalign'][0]).inv().q,
```
taldcroft, line 213: It's probably worth writing this out a bit, this expression is really difficult for me to parse.
```
# Transpose of transform/rotation matrix is the inverse
aca_mis_inv = acal['aca_misalign'][0].transpose()
# Repeat the transform to match asol
aca_mis_inv = np.repeat(aca_mis_inv[np.newaxis, ...], repeats=len(asol), axis=0)  
q_mis_inv = Quat(transform=aca_mis_inv)
```

mica/archive/asp_l1.py
```diff
@@ -210,13 +210,13 @@ def get_atts_from_files(asol_files, acal_files, aqual_files, filter=True):
...
                               len(asol)).reshape((4, len(asol))).transpose()
         # Quaternion multiply the asol quats with that inv misalign and save
         # I could also do this with the transform matrix and then only need
         # one accessory quat function.
         q_att_name = 'q_att_raw' if 'q_att_raw' in asol.colnames else 'q_att'
-        att_chunks.append((Quat(asol[q_att_name]) * Quat(q_mis_inv)).q)
+        att_chunks.append((Quat(q=asol[q_att_name]) * Quat(q=q_mis_inv)).q)
```
taldcroft, line 219: This looks suspicious in multiplying a `Quat` by an np.array `Quat.q`.  Clearly it was working with the mica `Quat`, but I think not for Quaternion.

taldcroft (commented):
LGTM apart from the comments

jeanconn:
Also WIP

mica/archive/asp_l1.py
javierggt: that might be a good catch. But something like this should work, so I would argue that if it doesn't, perhaps it should be fixed.

javierggt: actually... on second read, I might not be understanding Tom's comment...

jeanconn: I think I evaluated the product and it looks good but I'll re-check.  That was part of my WIP.

I also was planning to look at the unit test coverage and what is hit here.

javierggt: this should work. If it doesn't, I think it's a Quat issue. The main thing is that the broadcasting rules should apply to the Quat class. So a scalar times an array should be fine.

taldcroft: I think that the current behavior in Quaternion is correct, namely multiplying a `Quat` * `ndarray` should fail. However, I see now that I had mis-parsed the parens and thought that the operation was `q1 * q2.q` but in fact it is `(q1 * q2).q`.
```
In [2]: q = Quat([1,0,0,0])                                                                                                                       

In [3]: q.equatorial                                                                                                                              
Out[3]: array([  0.,   0., 180.])

In [4]: q * q                                                                                                                                     
Out[4]: <Quat q1=-0.00000000 q2=-0.00000000 q3=-0.00000000 q4=1.00000000>

In [5]: q * q.q                                                                                                                                   
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-5-1a633dcf67b3> in <module>
----> 1 q * q.q

~/miniconda3/envs/ska3-shiny/lib/python3.8/site-packages/Quaternion/Quaternion.py in __mul__(self, quat2)
    641         """
    642         q1 = np.atleast_2d(self.q)
--> 643         q2 = np.atleast_2d(quat2.q)
    644         assert q1.shape == q2.shape
    645         mult = np.zeros(q1.shape)

AttributeError: 'numpy.ndarray' object has no attribute 'q'
```

jeanconn: So with regard to the mica code that uses Quaternion, am I free to just use whatever works?

jeanconn:
In mica, with these changes, I think we're just left with a call to mica.quaterion in the mica.web.pcad_table app.  I think that's still run from the Python 2 kadi server?  I don't know when we'd plan to update that kind of thing or pull out of mica and put in a web-apps area.

mica/archive/asp_l1.py
jeanconn: It looks like this part of the code is hit by the mica.archive.tests.test_asp_l1.test_get_atts_time, which looks to be a silly test that gets a few days of L1 asols and compares to OBC quats.  The test is still good after this change so I think I've used something that works.  I suppose here in get_atts_from_files maybe we also need a comment that this transformation is about getting out of MNC frame (IIRC)?

javierggt: Regarding the error @taldcroft shows. I think that, at the very least, it is hard to figure out.

There are two possibilities, and I think both are fine:
- implicitly cast the right-side into a Quat. i.e.: doing
  ```python
  quat2 = Quat(attitude=quat2)
  q2 = np.atleast_2d(quat2.q)
  ```
- explicitly check the type so one gets the proper error:
   ```python
   assert isinstance(quat2, Quat), 'right-hand is not a Quat instance'
   ```

or a combination of both.

mica/archive/asp_l1.py
jeanconn: Updated to be written out per Tom's comment.

mica/quaternion.py
```diff
@@ -101,6 +102,7 @@ class Quat(object):
 
     """
     def __init__(self, attitude, intype=None):
+        warnings.warn("mica.quaternion is deprecated as of Quaternion 3.5.0", FutureWarning)
```
jeanconn, line 105: I don't know if adding the warning just causes more trouble for ourselves, but seems fine in this application.

taldcroft (approved):
LGTM

PR #242, opened 2020-09-25
Remove warning when account doesn't have axafvv access

taldcroft:
## Description

Small change in one test so that automated Ska3 regression testing can pass running as jgonzale.

## Testing

- [x] Passes unit tests on MacOS
- [N/A] Functional testing


jeanconn:
I am of the opinion that the account warning should be ignored on the ska_testr side.

jeanconn:
Well, and jgonzale should get the access.

taldcroft:
> I am of the opinion that the account warning should be ignored on the ska_testr side.

It is actually not that easy to ignore warnings like this in ska_testr. Instead you need to change `testr` : https://github.com/sot/testr/blob/master/testr/runner.py#L12. The problem is that pytest emits the whole warning section in addition to the particular line.

Once you tell `testr` to ignore it then `python setup.py test` will ignore it, and then there is not really much point in emitting the warning to begin with.

I guess if it is not acceptable to routinely skip this test (if run without credentials) then we need to ensure that @javierggt has credentials.

taldcroft:
Thinking more, this test relates to a single-user (@jeanconn) function used only in automated processing. If it ever fails then we'll get warnings / errors in the logs. I mostly just want to get ska_testr clean at this point.

PR #245, opened 2020-09-29
removed outdated documentation section

javierggt:
## Description

Strictly speaking, this should be replaced by a new section with the right information. I just don't know what that is and want to get rid of the warnings.

## Testing

- [ ] Passes unit tests on MacOS, linux, Windows (at least one required)
- [ ] Functional testing

Fixes #


jeanconn (approved):
This looks fine to me.  If we want to add some breadcrumbs about acdc promotion we could but I don't see a fire under that.

PR #248, opened 2020-11-09
Fixes for asp1 processing

jeanconn:
## Description

Fixes for asp1 processing.
- [x] trim bigger files from the asp1 download (kalm1, gdat1, adat1)
- [x] remove unnecessary gdat use in mica.vv
- [x] fix a bug in aspect_1_proc that wasn't updating the isdefault status

## Testing

- [x] Passes unit tests on Linux
- [x] Functional testing 

Description of functional testing:

I ran "on the side" in rebuild mode to confirm that it downloads and then deletes the kalm1, gdat, adat files.
I tested the aspect proc update bug on a local processing db3 by setting an incorrect status for one obsid/revision, running the update() method and confirming that the isdefault status was correctly set from the archfiles reference.

Fixes #


taldcroft (approved):
I have not looked at the code, but the concept and testing is good.

PR #249, opened 2020-11-12
More centroid_dashboard tweaks

jeanconn:
## Description

A collection of fiddly little commits to: reduce extra processing, reduce skips or error-out observations, update page html to avoid reflows.

## Testing

- [x] Functional testing

For functional testing, have just run a bunch of year 2020 with centroid dashboard code.  It is still filling in 

https://icxc.cfa.harvard.edu/aspect/centroid_reports

Fixes #

### DIFF
--- a/mica/archive/asp_l1.py
+++ b/mica/archive/asp_l1.py
@@ -220,7 +220,10 @@ def get_atts_from_files(asol_files, acal_files, aqual_files, filter=True):
         att_chunks.append((Quat(asol[q_att_name]) * Quat(q_mis_inv)).q)
         time_chunks.append(np.array(asol['time']))
         records.append(asol.meta)
-    return np.vstack(att_chunks), np.hstack(time_chunks), records
+    if len(att_chunks) > 0 and len(time_chunks) > 0:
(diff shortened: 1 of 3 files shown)


jeanconn:
This isn't needed for shiny, I just had it lingering.

taldcroft:
Excellent! One small request while you are at it is to change the y-axis label on the delta roll plot from `dr (arcsec)` to `Roll error (arcsec)` or (if that doesn't fit) `Roll err (arcsec)`.  `dr` is ambiguous.

taldcroft:
Another feature request - CCD temperature. This could be put in the Roll error plot window as a right-side axis.

taldcroft:
In the Performance details page, can you change Ending roll error to be "Not available" for ER's instead of 0.0? Not a huge deal but the 0.0 is wrong.

taldcroft:
Can you put the Catalog and Visualization box on the right side in a 3rd column? This would optimize for modern wide screens and let me see nearly the entire page at once when flipping through.

jeanconn:
I'm going to merge this and take those little RFEs and do them in a new PR.  Thanks @taldcroft !

PR #251, opened 2020-11-24
Add list of available content types to asp_l1.get_files

taldcroft:
## Description

I could never remember ACACENT, exactly.

## Testing

- [N/A] Passes unit tests on MacOS, linux, Windows (at least one required)
- [N/A] Functional testing


jeanconn (commented):
For the new / repro observations there's also ASPSOLOBI .

taldcroft:
Fixed.  What is ASPSOLOBI?

jeanconn:
ASPSOLOBI should be the combined aspect solution for an obi.  Introduced as part of the axis correction code.

PR #252, opened 2020-12-18
L0 ingest issues (both shiny and repro V)

jeanconn:
## Description

Fix some L0 ingest hiccups.  

The ingest code was failing to ingest single new files in Repro.  It looks like that was due to a bug in identifying new files that had been downloaded.  I also hit a db locking issue that was improved by reorganizing.  And while at it, I updated the pieces needed to make a new archive (of the list of files released.. the "cda" h5 piece) to work more happily with shiny (removed six, updated some urllib pieces).  I was not rigorous distinguishing reproV from shiny updates here as I have just been working in shiny.

## Testing

- [x] Passes mica archive l0 unit tests on Linux (a non-related L1 test looks to be failing now)
- [x] Functional testing - this worked in a test archive to ingest the files that were failing repeatedly with the flight code on the flight archive.  At this point I've also run the test code on the flight archive to just get the files in there.

### DIFF
--- a/mica/archive/aca_l0.py
+++ b/mica/archive/aca_l0.py
@@ -5,6 +5,7 @@ import re
 import logging
 import shutil
 import time
+import gzip
 from astropy.table import Table
 import astropy.io.fits as pyfits
 import numpy as np
@@ -752,7 +753,6 @@ class Updater(object):
             filename = have_files[0]
             # if it isn't gzipped, just gzip it
             if re.match(r'.*\.fits$', filename):
-                import gzip
                 f_in = open(file, 'rb')
                 f_out = gzip.open("%s.gz" % filename, 'wb')
                 f_out.writelines(f_in)
--- a/mica/archive/aca_l0.py
+++ b/mica/archive/aca_l0.py
@@ -40,7 +40,7 @@ FILETYPE = {'level': 'L0',
             'instrum': 'PCAD',
             'content': 'ACADATA',
             'arc5gl_query': 'ACA0',
-            'fileglob': '*fits.gz'}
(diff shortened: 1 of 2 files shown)


mica/archive/aca_l0.py
```diff
@@ -769,14 +765,14 @@ def _fetch_individual_files(self, files):
...
-                    self._move_archive_files([f])
+        for i, f in enumerate(files):
+            arch_info = self._read_archfile(i, f, files)
+            if arch_info:
+                self._move_archive_files([f])
+                with Ska.DBI.DBI(**self.db) as db:
                     db.insert(arch_info, 'archfiles')
-                    count_inserted += 1
-            db.commit()
+                    db.commit()
```
jeanconn, line 774: It is less efficient here to commit after each insert, but the "move_archive_files" routine needs a lock on the db and it can't do that if there are uncommited inserts.

mica/archive/aca_l0.py
```diff
@@ -739,7 +737,7 @@ def _fetch_individual_files(self, files):
         fetched_files = []
         ingest_dates = []
         # get the files, store in file archive, and record in database
-        for file in files:
+        for i, file in enumerate(files):
```
taldcroft, line 740: I don't see a dependence on the new `i`.

taldcroft (approved):
Superficial review, but if your testing is good then this is fine.

mica/archive/aca_l0.py
jeanconn: Good point.  I can cut that.  I was using it for extra debugging when troubleshooting.

PR #256, opened 2021-01-13
Skip report vv warn if no sybase passwd file for db

jeanconn:
## Description

Skip report vv warn if no sybase passwd file for db

## Testing

- [x] Functional testing
   - ran as jeanconn and kadi user on Linux from production ska3 and see test pass and test skip as appropriate

Fixes #

### DIFF
--- a/mica/report/tests/test_write_report.py
+++ b/mica/report/tests/test_write_report.py
@@ -9,15 +9,20 @@ from testr.test_helper import on_head_network, has_sybase
 
 from .. import report
 
+user = os.environ.get('USER') or os.environ.get('LOGNAME')
+
 try:
     import Ska.DBI
-    user = os.environ.get('USER') or os.environ.get('LOGNAME')
     with Ska.DBI.DBI(server='sqlsao', dbi='sybase', user=user, database='axafvv') as db:
         HAS_SYBASE_ACCESS = True
 except Exception:
-    if on_head_network() and not has_sybase():
-        warn("On HEAD but no sybase access. Run ska_envs or define SYBASE/SYBASE_OCS")
     HAS_SYBASE_ACCESS = False
+
+    # If the user should have access, warn about the issue.
+    if (on_head_network() and not has_sybase() and
+            os.path.exists(os.path.join(os.environ['SKA'], 'data', 'aspect_authorization',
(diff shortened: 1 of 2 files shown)


jeanconn:
More klunky compromise here, but I think this keeps me happy about test coverage and could stop the warns (though now @javierggt would need to *delete* his passwd file if it isn't actually working for test to pass... and put it back when it works).

mica/report/tests/test_write_report.py
```diff
@@ -9,16 +9,21 @@
 
 from .. import report
 
+user = os.environ.get('USER') or os.environ.get('LOGNAME')
```
taldcroft, line 12: I think the best way to get user information is this, which you can reliably use on any system:
```
import getpass
getpass.getuser()
```

mica/report/tests/test_write_report.py
```diff
@@ -9,16 +9,21 @@
...
     with Ska.DBI.DBI(server='sqlsao', dbi='sybase', user=user, database='axafvv') as db:
         HAS_SYBASE_ACCESS = True
 except Exception:
-    if on_head_network() and not has_sybase():
-        warn("On HEAD but no sybase access. Run ska_envs or define SYBASE/SYBASE_OCS")
     HAS_SYBASE_ACCESS = False
 
+    # If the user should have access, warn about the issue.
+    if (on_head_network() and not has_sybase() and
+            os.path.exists(os.path.join(os.environ['SKA'], 'data', 'aspect_authorization',
```
taldcroft, line 23: You should just start using `Path` at every opportunity. It is very much nicer in almost every case:
`Path(os.environ['SKA'], 'data', 'aspect_authorization', f'sqlsao-axafvv-{user}').exists()`.

taldcroft (changes requested):
Both of the changes are not required, but are about learning (through usage) new and better patterns for doing common things.

jeanconn:
Sure.  Yes, I was just extending what was there (at least in my mind) instead of doing it a little better.  I can do it a little better.

jeanconn:
@javierggt up to you if you want to leave your sqlsao-axafvv-* file around or if you want to delete so the tests will pass against master until vv access actually works.

PR #257, opened 2021-01-24
Avoid vv obspar race condition

jeanconn:
## Description

Avoid vv obspar race condition.

With repro going on, we're getting a few V&V hiccups.  It looks like most are due to asp L1 products getting ingested in mica before obspars are available.  These PR changes 1) set the obspar ingest to run after the L1 ingest in the task schedule, 2) make the processing more robust to a missing obspar (log and skip).

## Testing

- [ ] Functional testing

I did not put time into creating a good functional test.  As this is a race condition situation, one would need to construct a state in the various databases matching the mismatch seen at the time of the V&V hiccup.  Given the benign nature of the changes, I think testing that this fixes the problem by just running the code in flight (and looking for warnings) makes sense.

Fixes #

### DIFF
--- a/mica/vv/process.py
+++ b/mica/vv/process.py
@@ -109,43 +109,33 @@ def update(obsids=[]):
 
     :param obsids: optional list of obsids
     """
-    # if an obsid is requested, just do that
-    # there's a little bit of duplication in this block
-    if len(obsids):
-        for obsid in obsids:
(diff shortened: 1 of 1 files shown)


taldcroft (approved):
I can't say that I have scrutinized the code changes, but the idea in the description sounds fine, along with justification for no functional testing.

PR #258, opened 2021-04-01
Disable use of agasc supplement for stats and a specific vv use case

jeanconn:
## Description

Disable use of agasc supplement for stats and a specific vv use case
Also update a few vv tests newly failing due to repro.

## Testing

- [ ] Functional testing


jeanconn:
I haven't run the stats "on-the-side" to test this so it has seen no functional testing.  Given the "use_supplement" scope, I think we just need to be sure I didn't spell the option incorrectly and we're good.

taldcroft (approved):
LGTM

PR #259, opened 2021-05-03
Add get_star_stats single-star method for acq/gui stats and remove Sybase from mica.web.star_hist

jeanconn:
## Description

Add some methods to do single star queries to the stats fetchers and update the mica.web star history piece to not use Sybase for guide stats.

## Testing

- [x] Passes unit tests on MacOS
- [x] Functional testing - two silly read-tests added for new methods.  

Not fully integration tested in kadi just yet.

### DIFF
--- a/mica/stats/acq_stats.py
+++ b/mica/stats/acq_stats.py
@@ -13,11 +13,10 @@ def get_stats(filter=True):
     :returns acq_stats: numpy.ndarray
     """
 
-    h5 = tables.open_file(TABLE_FILE, 'r')
-    stats = h5.root.data[:]
-    h5.close()
-    if filter:
-        stats = stats[~stats['known_bad']]
+    with tables.open_file(TABLE_FILE, 'r') as h5:
+        stats = h5.root.data[:]
+        if filter:
+            stats = stats[~stats['known_bad']]
     return stats
 
 
@@ -29,8 +28,8 @@ def get_star_stats(id, filter=True):
     :param filter: True filters out rows marked 'known_bad' in table
     :returns acq_stats: numpy.ndarray
     """
-    h5 = tables.open_file(TABLE_FILE, 'r')
-    stats = h5.root.data.read_where(f'agasc_id == {id}')
-    if filter:
-        stats = stats[~stats['known_bad']]
+    with tables.open_file(TABLE_FILE, 'r') as h5:
+        stats = h5.root.data.read_where(f'agasc_id == {id}')
+        if filter:
+            stats = stats[~stats['known_bad']]
     return stats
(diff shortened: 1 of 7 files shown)


jeanconn:
These were tiny changes for mica web to not use sybase.  Added a couple of other convenience methods.  Seems it could just go but no rush.

mica/stats/acq_stats.py
```diff
@@ -19,3 +19,18 @@ def get_stats(filter=True):
...
+
+def get_star_stats(id, filter=True):
+    """
+    Retrieve acq stats for agasc id
+
+    :param id: agasc id
+    :param filter: True filters out rows marked 'known_bad' in table
+    :returns acq_stats: numpy.ndarray
+    """
+    h5 = tables.open_file(TABLE_FILE, 'r')
```
taldcroft, line 32: Use a context manager, you are not explicitly closing `h5`.

mica/stats/guide_stats.py
```diff
@@ -21,3 +21,16 @@ def get_stats(filter=True):
...
 
+def get_star_stats(id, filter=True):
+    """
+    Retrieve guide stats for agasc id
+
+    :param id: agasc id
+    :param filter: True filters out rows marked 'known_bad' in table
+    :returns stats: numpy.ndarray
+    """
+    h5 = tables.open_file(TABLE_FILE, 'r')
```
taldcroft, line 32: Context manager.

mica/stats/tests/test_acq_stats.py
```diff
@@ -2,11 +2,21 @@
...
 
 from .. import update_acq_stats as acq_stats
+from .. import acq_stats as read_acq_stats
 
 HAS_OBSPAR_ARCHIVE = os.path.exists(
         acq_stats.mica.archive.obspar.CONFIG['data_root'])
+HAS_ACQ_TABLE = Path(read_acq_stats.TABLE_FILE).exists()
+
+
+@pytest.mark.skipif('not HAS_ACQ_TABLE', reason='Test requires acq stats table')
```
taldcroft, line 15: Modern preferred idiom is to use the value, not the string repr. So just `skipif(not HAS_ACQ_TABLE, ..)`.

taldcroft (commented):
Looks good with some small changes. Can you update the PR title to reflect that it also adds the `get_star_stats` methods for acq and guide? I went into the review expecting mostly changes about sybase.

jeanconn:
Thanks Tom!  Also, for the star history app, @javierggt where should I be linking for the supplement star reports?  This PR was the minimal set that occurred to me to remove Sybase but I do want to both report the supplement mags and link into the reports if available.

javierggt:
as far as where to find the supplement star reports, I would do something like

```python
url = f'https://cxc.cfa.harvard.edu/mta/ASPECT/agasc/supplement_reports/stars/{int(agasc_id//1e7):03d}/{agasc_id}/index.html'
```

taldcroft (approved):

PR #262, opened 2021-09-01
Skip tests if obspar data dirs not present

jeanconn:
## Description

Update the "skip if no obspars" test for test skipping to skip if the '00' directory isn't present within the obspars directory.  The obspars directory with one database file is now sync'd as part of ska_sync but the data for the tests in the subdirectories isn't there.

## Testing

- [x] Passes unit tests on chimchim and skips the acq and guide tests correctly.
- [x] Functional testing

In this case the functional testing was the unit testing.


jeanconn:
I didn't make anything prettier but I think this at least stops the newer mica fails on GRETA.

javierggt (approved):
LGTM

PR #264, opened 2021-09-20
Make a target table product

jeanconn:
## Description

 Make a target table product

## Testing

- [x] Functional testing - added tiny read/write unit test as functional test


docs/ocat_target_table.rst
```diff
@@ -0,0 +1,157 @@
...
+MULTITEL_INT
+CONSTR_RMK                    Constraint in remarks flag
+TOO_TYPE
+TOO_START
+TOO_STOP
+ALT_GROUP
+ALT_TRIG
+SIMODE                        Science Instrument (SI) mode
+HRC
+SPECT_MODE
```
taldcroft, line 114: ??

mica/archive/ocat_target_table.py
```diff
@@ -0,0 +1,28 @@
...
+    :param datafile: h5 ocat target table data file.
+                     Defaults to MICA_ARCHIVE/ocat_target_table.h5
+    :returns: astropy table of target table
+    """
+    if datafile is None:
+        datafile = OCAT_TARGET_TABLE
+
+    dat = Table.read(datafile)
+
+    # Decode bytes to strings manually.  Perhaps fixed in newer versions of numpy.
```
taldcroft, line 22: Definitely fixed in newer versions of numpy.

docs/ocat_target_table.rst
jeanconn: Not a misspelling of ASPECT.  I assume this is some HRC thing but I'm not that HRC fluent.
```
In [10]: np.unique(dat['SPECT_MODE'])                                           
Out[10]: 
<Column name='SPECT_MODE' dtype='str12' length=4>
       (DEF)
           0
     IMAGING
NORMAL (DEF)
```

mica/archive/ocat_target_table.py
jeanconn: Hah.  The "perhaps" was more meant to suggest that I have no idea what we'll use in ska3, but fair.

jeanconn:
Let's lower-case all the column names.

mica/archive/update_ocat_target_table.py
```diff
@@ -0,0 +1,89 @@
...
+    # Convert the type line to standard RDB
+    # First find the line that begins the column descriptions
+    for i, line in enumerate(lines):
+        if line.startswith('SEQ_NUM'):
+            header_start = i
+            break
+
+    # The line with the lengths and types is next (header_start + 1)
+    ctypes = lines[header_start + 1].split("\t")
+    for j, t in enumerate(ctypes):
```
taldcroft, line 69: Avoid using single letter, non-descriptive variables for anything but counters like `i`, `j`.  Here `ctype` would be a good choice.

mica/archive/update_ocat_target_table.py
```diff
@@ -0,0 +1,89 @@
...
+    lines = resp.text.split("\n")
+
+    # Convert the type line to standard RDB
+    # First find the line that begins the column descriptions
+    for i, line in enumerate(lines):
+        if line.startswith('SEQ_NUM'):
+            header_start = i
+            break
+
+    # The line with the lengths and types is next (header_start + 1)
```
taldcroft, line 67: Try to make comments descriptive of what the code is doing and *why*. Like how the RDB standard requires strings types to be something like `20S` but this file uses `20`, so munge things accordingly.

mica/archive/update_ocat_target_table.py
```diff
@@ -0,0 +1,89 @@
...
+
+def main():
+    opt = get_options()
+    datafile = Path(opt.datafile)
+    update_table(datafile)
+
+
+def update_table(datafile):
+    table = get_web_table()
+    if datafile.name.endswith('h5'):
```
taldcroft, line 33:
```suggestion
    if datafile.suffix == '.h5':
```

mica/archive/update_ocat_target_table.py
```diff
@@ -0,0 +1,89 @@
...
+        description="Update target table")
+    parser.add_argument("--datafile",
+                        default='target_table.h5')
+    opt = parser.parse_args()
+    return opt
+
+
+def main():
+    opt = get_options()
+    datafile = Path(opt.datafile)
```
taldcroft, line 27: Better to put data type munging into the lower level function, that way someone could pass a string to `update_table` and it would work.

mica/archive/update_ocat_target_table.py
```diff
@@ -0,0 +1,89 @@
...
+import argparse
+from pathlib import Path
+from astropy.table import Table
+import requests
+import numpy as np
+
+from ska_helpers.retry import retry_call
+from mica.common import MICA_ARCHIVE
+
+URL_OCATDETAILS = "https://cda.harvard.edu/srservices/ocatDetails.do?format=text"
```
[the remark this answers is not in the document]
jeanconn: Yes I got details on the API (at the local file I'd shared in one of our chats... figure I shouldn't put that here).  I figured if we've got a plan to just update a local copy of the table we don't need to expose the other parts of the API.  If you want to do so it seems like we'd move that to the more public module etc.

I think you can search any of the columns/parameters
```
There are four kinds of parameters:
comma-delimited lists (e.g. instrument=ACIS-I,ACIS-S,HRC-I,HRC-S)
single value from list (e.g. inputCoordFrame=J2000 -- other options: b1950, bxxxx, ec1950, ecxxxx, galactic)
single value (e.g. target=
range lists (hyphen-separated, e.g. obsid: obsid=100,200-300,600-1000,1800)
```
There's also a note that there are some other pieces (acis windows, roll requirements) that can be retrieved if not a text query (text looks like it needs to be a flat return table).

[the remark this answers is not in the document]
taldcroft: > at the local file I'd shared in one of our chats

Can you remind me? Actually my biggest motivation is to make the test not take so long. Can you just get a handful of obsids?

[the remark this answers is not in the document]
jeanconn: Oh sure.  I'll send you the file.  But I can also just make the test just get a handful of obsids.  Mostly I had the test do what the regular job *would* do as I figured it didn't take long, the mica tests already take too long, and then we were testing just what we'd be using.  But fine to make a shorter test (and fix the rest of the tests at some point).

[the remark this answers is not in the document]
taldcroft: Anyway, you can just leave it for now. I have some ideas for enhancements, but let's just get this merged mostly as-is, just fixing the real problems I put into comments.

docs/ocat_target_table.rst
```diff
@@ -0,0 +1,157 @@
...
+OC_COL
+EVFIL
+EVFIL_LO
+EVFIL_RA
+EFFICIENT                     ACIS use most efficient (Y, N)
+SPWIN                         Spatial window (Y, N)
+============================= ================================================================
+
+The hdf5 in-kernel searches may be faster working with the table directly for some
+operations.
```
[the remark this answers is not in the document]
jeanconn: Sure.  Seems fine but I note that including read_where seems to throw a bunch of these warnings
```
mica/archive/tests/test_target_table.py::test_write_read_table
  /Users/jean/miniconda3/envs/ska3/lib/python3.8/site-packages/tables/path.py:155: NaturalNameWarning: object name is not a valid Python identifier: 'mon_max.mask'; it does not match the pattern ``^[a-zA-Z_][a-zA-Z0-9_]*$``; you will not be able to use natural naming to access this object; using ``getattr()`` will still work, though
    check_attribute_name(name)
```
Not sure if we want to shut those up or ?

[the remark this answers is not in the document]
taldcroft: Well never mind for now. Let's go for the incremental development and just get something in.

mica/archive/tests/test_target_table.py
```diff
@@ -0,0 +1,23 @@
...
+def test_write_read_table():
+    """
+    Make a report and database
+    """
+    tempdir = tempfile.mkdtemp()
+    file = Path(tempdir, 'target_table.h5')
+    update_ocat_target_table.update_table(datafile=file)
+    dat = ocat_target_table.get_ocat_target_table(datafile=file)
+    ok = dat['obsid'] == 2121
+    dat['target_name'][ok][0] == "MCG-5-23-16"
```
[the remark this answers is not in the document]
jeanconn: Thanks!

mica/archive/update_ocat_target_table.py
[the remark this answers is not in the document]
jeanconn: OK.  I just stuck in a `filter` option but if that is poorly named or should be cut that's fine too.

mica/archive/update_ocat_target_table.py
jeanconn: I added the comment you wanted.  I think this one will also be useful for a future coder if it comes up though.

taldcroft (approved):
I'm planning on making a few changes, but we can go with this now. (I just want to hack on some fun little thing instead of the endless manager tasks of late).

jeanconn:
I'm not actually seeing which piece you are meaning for "never mind for now".  Do you mean you want to pull out the `read_where` as a bit problematic?  It still works.

taldcroft:
I think messages got crossed up in threads. All good.

PR #267, opened 2021-11-15
Add CDA web services API and expand functionality of Ocat target target query

taldcroft:
## Description

Expanding on #264, this makes a somewhat unified interface for directly querying the CDA web services and querying from a local HDF5 table. I noticed the CDA web interface is reasonably good and thought it would be worth a Python interface so people don't necessarily have to sync the mica target table archive file. This is also the result of too much management and wanting to do a fun hack.

Features:
- Puts things in a new `cda` subpackage (along with `aca0_list.py`)
- It turns out the in-kernel HDF5 is quite fast, so this takes advantage of that to the extent possible.
- Converts a result from a single obsid from a 1-row table to a dict (slightly awkward in the docs, but this is probably what the user wants in most cases).
- Unified target_name search with option for resolving the name into coordinates
- Unified search by coordinate
- Applies units
- Includes the available CDA search parameters in the docs.

ToDo:
- [x] Docs
- [x] Testing
- [x] Cache full local HDF5 table with a timeout (e.g. an hour)

## Testing

- [x] Functional testing

### Functional testing

Installed locally (`python setup.py install --user`) and confirmed proper creation of the local HDF5 with:
```
~/.local/bin/mica_update_ocat_target_table --datafile=ocat.h5
```

Also modified the code to have a short timeout (1 sec) and confirmed that it tries 3 times and then gives up.
QUESTION: maybe it should "succeed" but print an error message to console to prevent stopping other jobs?

Built the docs locally with no warnings and they seem OK.


mica/archive/cda/services.py
```diff
@@ -0,0 +1,665 @@
...
+        Returned text from the service
+    """
+
+    if service not in CDA_SERVICES:
+        raise ValueError(f'unknown service {service!r}, must be one of {list(CDA_SERVICES)}')
+
+    # Query the service and check for errors
+    url = f'{URL_CDA_SERVICES}/{CDA_SERVICES[service]}.do'
+    verbose = params.pop('verbose', False)
+    print(f'{timeout=}')
```
jeanconn, line 402: I think this is a lingering debug piece.

docs/cda_ocat.rst
```diff
@@ -0,0 +1,296 @@
...
+least 10 times faster than using the `Web access to CDA`_.
+
+The local version of the Ocat is cached in memory the first time it is read in
+full, so this means subsequent full reads or searches for target names will be
+much faster. The cache expires in one hour in case you have a long-running job.
+
+Any of the fields in the returned table can be used for filtering the query
+via exact matches of the parameters::
+
+   >>> from mica.archive.cda import get_ocat_local
```
jeanconn, line 48: I think there was some function renaming after the very comprehensive docs?  Because this doesn't work for me.

mica/archive/cda/services.py
```diff
@@ -0,0 +1,665 @@
...
+      AND the obsid corresponds to an integer
+      AND the returned result has a single row
+      THEN the return type is ``dict``
+      ELSE the return tuple is a ``Table``.
+
+    If ``return_type='table'`` then always return a ``Table``."""
+
+CDA_PARAM_DOCS = """Additional function args for CDA search parameters::
+
+        instrument=ACIS,ACIS-I,ACIS-S,HRC,HRC-I,HRC-S
```
jeanconn, line 63: I'm not sure where the most useful place is to make it clear that the local search and the non-local search are against different column names?

docs/cda_ocat.rst
jeanconn: Oops, thought I was working in a brand new checkout of the branch but it was stale.

PR #271, opened 2022-05-09
Use CxoTime and fix time range for update

jeanconn:
## Description

Use CxoTime instead of DateTime and fix update() in mica.report.  As of kadi 5.9.1 get_states
called in update() cannot take "-7" as a relative time from now.

## Interface impacts
None

### Unit tests

- [x] Linux

Independent check of unit tests by @taldcroft
- [x] Linux

### Functional tests

I manually confirmed that the CxoTime calls give equivalent output to the DateTime calls.

### DIFF
--- a/mica/report/report.py
+++ b/mica/report/report.py
@@ -22,7 +22,7 @@ import astropy.units as u
 from Ska.Shell import bash
 import agasc
 import Ska.DBI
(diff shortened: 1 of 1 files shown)


taldcroft (approved):
LGTM

Issue #268, opened 2022-01-10
dark cal methods don't handle bounds


jeanconn:
If one asks for a dark cal before the earliest one in the archive, one just gets the latest one.  An error would be more appropriate I think.

```
In [1]: from mica.archive import aca_dark                                                                                                              

In [2]: aca_dark.dark_cal.get_dark_cal_props(date='2000:326', select='before')['date']                                                                 
Out[2]: '2000:326'

In [3]: aca_dark.dark_cal.get_dark_cal_props(date='2000:325', select='before')['date']                                                                 
Out[3]: '2021:329:10:10:29.450'
```

PR #273, opened 2022-06-09
Added vectorized versions of some archive.dark_cal functions

javierggt:
## Description

This PR adds vectorized versions of get_dark_cal_id, date_to_dark_id and dark_id_to_date. It also replaces the uses of DateTime by CxoTime inside `mica.archive.aca_dark.dark_cal` (all other uses of `DateTime` were left there).

## Interface impacts
Everything *should* be backward compatible. This only enhances the interface, so something like this is possible:
```
get_dark_cal_id(['2022:100', '2022:105', '2022:127'], 'before')
```

### Unit tests

I added a test for these vectorized functions.

- [x] Mac

Independent check of unit tests by @jeanconn 
- [x] Linux

### Functional tests

No functional testing.

### DIFF
--- a/mica/archive/aca_dark/dark_cal.py
+++ b/mica/archive/aca_dark/dark_cal.py
@@ -68,7 +68,7 @@ def get_dark_cal_dirs(dark_cals_dir=MICA_FILES['dark_cals_dir'].abs):
     Get an ordered dict of directory paths containing dark current calibration files,
     where the key is the dark cal identifier (YYYYDOY) and the value is the path.
 
-    :param source: source of dark cal directories ('mica'|'ska')
+    :param dark_cals_dir: directory containing dark cals.
     :returns: ordered dict of absolute directory paths
     """
     dark_cal_ids = sorted([fn for fn in os.listdir(dark_cals_dir)
@@ -76,6 +76,20 @@ def get_dark_cal_dirs(dark_cals_dir=MICA_FILES['dark_cals_dir'].abs):
     dark_cal_dirs = [os.path.join(dark_cals_dir, id_)
                      for id_ in dark_cal_ids]
     return OrderedDict(zip(dark_cal_ids, dark_cal_dirs))
+
+
+@lru_cache()
+def get_dark_cal_ids(dark_cals_dir=MICA_FILES['dark_cals_dir'].abs):
+    """
+    Get an ordered dict dates as keys and dark cal identifiers (YYYYDOY) as values.
+
+    :param dark_cals_dir: directory containing dark cals.
+    :returns: ordered dict of absolute directory paths
+    """
+    dark_cal_ids = sorted([fn for fn in os.listdir(dark_cals_dir)
+                           if re.match(r'[12]\d{6}$', fn)])
+    dates = [CxoTime(d[:4] + ':' + d[4:]).date for d in dark_cal_ids]
+    return OrderedDict(zip(dates, dark_cal_ids))
 
 
 def get_dark_cal_id(date, select='before'):
@@ -118,7 +132,7 @@ def _get_dark_cal_id_scalar(date, select='before'):
     if ii < 0:
         earliest = CxoTime(dark_cal_secs[0]).date[:8]
         raise ValueError(
-            f'trying to get last dark cal before {date}, '
+            f'trying to get last dark cal on {date}, '
             f'which is before the earliest dark cal on {earliest}'
         )
 
--- a/mica/archive/aca_dark/dark_cal.py
+++ b/mica/archive/aca_dark/dark_cal.py
@@ -115,6 +115,13 @@ def _get_dark_cal_id_scalar(date, select='before'):
     else:
         raise ValueError('select arg must be one of "nearest", "before", or "after"')
 
+    if ii < 0:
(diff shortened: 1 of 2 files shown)


mica/archive/aca_dark/dark_cal.py
```diff
@@ -30,21 +30,34 @@ def date_to_dark_id(date):
     """
     Convert ``date`` to the corresponding YYYYDOY format for a dark cal identifiers.
 
-    :param date: any DateTime compatible format
+    :param date: any CxoTime compatible format
     :returns: dark id (YYYYDOY)
     """
-    date = DateTime(date).date
-    return date[:4] + date[5:8]
+    t = CxoTime(date)
```
taldcroft, line 36: No 1-letter variable names in new code. You can use `tm` as a standard "time", but in general always use descriptive variable names.

mica/archive/aca_dark/dark_cal.py
```diff
@@ -30,21 +30,34 @@ def date_to_dark_id(date):
...
+    shape = np.shape(t)
+    if not shape:
+        return t.date[:4] + t.date[5:8]
+    b = t.date.view((str, 1)).reshape(-1, 21)
```
taldcroft, line 40: You can use `t_date.dtype.itemsize // 4` to avoid that hardcoded 21. An alternate implementation inspired by similar code in astropy is:
```
dates = np.array(['2022:001:00:01:02.123']*6).reshape(2,3)
chars = dates.view((np.uint32, dates.dtype.itemsize // 4))
ids_char = np.concatenate([chars[..., :4], chars[..., 5:8]], axis=-1)
dark_ids = ids_char.view((str, dark_ids.shape[-1])).squeeze(-1)
```

taldcroft (commented):
This looks like a fine idea. 

:+1: always on using `CxoTime` instead of `DateTime`.

mica/archive/aca_dark/dark_cal.py
```diff
@@ -11,10 +11,11 @@
 from astropy.io import fits, ascii
 from astropy.table import Column
 import pyyaks.context
-from Chandra.Time import DateTime
+from cxotime import CxoTime
 from chandra_aca.aca_image import ACAImage
 
-from chandra_aca.dark_model import dark_temp_scale, DARK_SCALE_4C
+from chandra_aca.dark_model import dark_temp_scale
+from chandra_aca.dark_model import DARK_SCALE_4C  # noqa
```
jeanconn, line 18: Is this a new convention or?

mica/archive/aca_dark/dark_cal.py
```diff
@@ -30,21 +31,35 @@ def date_to_dark_id(date):
...
     :returns: dark id (YYYYDOY)
     """
-    date = DateTime(date).date
-    return date[:4] + date[5:8]
+    time = CxoTime(date)
+    date_str = time.date
+    if not time.shape:
+        return date_str[:4] + date_str[5:8]
+    date_str = np.atleast_1d(date_str)
+    b = date_str.view((str, 1)).reshape(-1, date_str.dtype.itemsize // 4)
```
jeanconn, line 42: I think these don't pass Tom's "no 1-letter variable names" feedback in this PR.

mica/archive/aca_dark/dark_cal.py
javierggt: I added the `# noqa` statement so flake doesn't complain about `DARK_SCALE_4C` being defined and not used, because I figured we still want to keep it so client code can use it as `dark_cal.DARK_SCALE_4C`.

I moved it on its own line so it is obvious what is it that flake is supposed to ignore. In other words, the import of dark_temp_scale should not be ignored by the linter, and if we ever remove the code that uses it flake should complain about the unused symbol.

jeanconn: Yet it was not obvious to me, as I wasn't paying attention to the original issue.... Not sure if we should leave it here for client code or not at this point then.  It is a good question.

jeanconn (approved):
LGTM. I note I think this doesn't fix https://github.com/sot/mica/issues/268 so now that bug is vectorized too.

mica/archive/tests/test_aca_dark_cal.py
```diff
@@ -104,3 +104,29 @@ def test_get_dark_cal_props_table_mixed():
...
+
+    dark_cal_id_ref = ['2022104', '2022104', '2022133']
+    dark_cal_id = dark_cal.get_dark_cal_id(['2022:100', '2022:105', '2022:127'], 'nearest')
+    assert np.all(dark_cal_id_ref == dark_cal_id)
+
+
+@pytest.mark.skipif('not HAS_DARK_ARCHIVE', reason='Test requires dark archive')
+def test_lower_limit():
+    with pytest.raises(ValueError, match='trying to get last dark cal on'):
+        dark_cal.get_dark_cal_id('2000:001')
```
jeanconn, line 132: This seems fine.  I don't know if we'll end up with a use case to put the 1999:223 OAC dark cal in the mica archive more directly.

jeanconn:
LGTM.  I reran the unit tests and re-approved.

PR #274, opened 2022-07-08
Fix issue in vectorized get_dark_cal_id

javierggt:
## Description

This fixes a bug found in regression testing. `get_dark_cal_id` is vectorized using `np.vectorize`. When the input is a scalar, it returns a numpy scalar, but that is a numpy array with shape `()`. This has overloaded the bracket operators, so you can't call the bracket operators on the result. The solution is to return a scalar of the dtype's type.

## Interface impacts
None

## Testing
mica unit tests and sparkles tests pass (including a new test for this)

### Unit tests

- [x] Mac

- [ ] [PLATFORM]:

### DIFF
--- a/mica/archive/aca_dark/dark_cal.py
+++ b/mica/archive/aca_dark/dark_cal.py
@@ -107,7 +107,7 @@ def get_dark_cal_id(date, select='before'):
     dark_id = _get_dark_cal_id_vector(date, select=select)
     if not dark_id.shape:
         # returning an instance of the type, not a numpy array
-        return dark_id.dtype.type(dark_id)
+        return dark_id.tolist()
     return dark_id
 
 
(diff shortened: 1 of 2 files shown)


mica/archive/aca_dark/dark_cal.py
```diff
@@ -104,7 +104,11 @@ def get_dark_cal_id(date, select='before'):
 
     :returns: dark cal id string (YYYYDOY)
     """
-    return _get_dark_cal_id_vector(date, select=select)
+    dark_id = _get_dark_cal_id_vector(date, select=select)
+    if not dark_id.shape:
+        # returning an instance of the type, not a numpy array
+        return dark_id.dtype.type(dark_id)
```
taldcroft, line 110: I think that `dark_id.tolist()` is a more standard idiom to get back to a pure Python representation. In this case it will return a `str` object instead of a `numpy.str_` object. Even though they are basically compatible they are still different object types.

javierggt: I changed it to use `tolist()`.

taldcroft (approved):

PR #276, opened 2022-08-25
add optional dark_cal_ids argument to dark_cal.get_dark_cal_id

javierggt:
## Description

This PR adds an optional argument to `dark_cal.get_dark_cal_id`. Normally one calls this function like this:
```
dark_cal.get_dark_cal_id(time, 'before')
```
and internally it checks all dark-cal IDs to see which one is "closer" according to the given selection criterion. After this change, one will be able to do
```
dark_cal_ids = dark_cal.get_dark_cal_ids()
dark_cal.get_dark_cal_id(time, 'before', dark_cal_ids=dark_cal_ids)
```

The reason for this is that I am modifying aca_view to fetch data from the Ska API, and I would like to reduce the number of calls to the API. In the first call above, one gets the dark-cal ID for a given time, and the call might be repeated many times. This is unnecessary  because in most cases there will only be one dark-cal ID that applies to all times.

I would rather get the whole list of dark-cal IDs from the API upfront, and then call the `get_dark_cal_id` method locally for all times.

### Unit tests

Unit tests were improved to include this new argument:

- [x] Mac

- [ ] [PLATFORM]:

### Functional tests

I have been using this from aca_view.

### DIFF
--- a/mica/archive/aca_dark/dark_cal.py
+++ b/mica/archive/aca_dark/dark_cal.py
@@ -92,7 +92,7 @@ def get_dark_cal_ids(dark_cals_dir=MICA_FILES['dark_cals_dir'].abs):
     return OrderedDict(zip(dates, dark_cal_ids))
 
 
-def get_dark_cal_id(date, select='before'):
+def get_dark_cal_id(date, select='before', dark_cal_ids=None):
     """
     Return the dark calibration id corresponding to ``date``.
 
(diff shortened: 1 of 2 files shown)


jeanconn:
"fetching a single dark-cal takes a whopping 1 minute. This is crazy."

Yikes.  Do you want to share the reduced snippet that is doing that for you?  I'm used to using something like this (this is on my laptop).

```
In [3]: %timeit mica.archive.aca_dark.get_dark_cal_image('2022:232')
11.3 ms ± 640 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [4]: %timeit mica.archive.aca_dark.get_dark_cal_id('2022:232')
4.07 ms ± 142 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

javierggt:
The issue is fetching a single dark-cal image. I can't imagine it has to take that long. I could do as you say and only fetch a section around the image, but that complicates things. In the end I might just say the dark-cal will not be fetched from the API.

jeanconn:
Yeah, for me the dark cal cutouts don't add much information so for the no-ska-data case just not including them seems better than working the issue.

taldcroft (commented):
Looks fine but needs documentation of some testing.

javierggt:
> Looks fine but needs documentation of some testing.

Done.

taldcroft (approved):

PR #277, opened 2022-09-01
starcheck subpackage: apply black and isort, fix line lengths, remove unused imports

taldcroft:
## Description

This tidies up the `mica.starcheck` subpackage in advance of removing the `timelines` database dependence.

## Interface impacts

None

### Unit tests

- [x] Mac

Independent check of unit tests by Jean
- [x] Linux

### Functional tests

The only non-automated check was fixing the regex expression in 26d18a4c108265998ead30b7a8827e86e73d90c0. (The forward slash `/` is not a special character and should not be escaped). I tested this with:
```
In [4]: st = '/data/mpcrit1/mplogs/2020/APR2020/oflsa/starcheck.txt'

In [5]:         mp_dir = re.search(
   ...:             r"/data/mpcrit1/mplogs(/\d{4}/\w{7}/\w{5}/)starcheck.txt", st
   ...:         ).group(1)
   ...: 

In [6]: mp_dir
Out[6]: '/2020/APR2020/oflsa/'
```

### DIFF
--- a/mica/starcheck/process.py
+++ b/mica/starcheck/process.py
@@ -160,7 +160,7 @@ def update(config=None):
         logger.info("Attempting ingest of %s" % st)
         # get an existing id or insert a new one
         mp_dir = re.search(
-            "\/data\/mpcrit1\/mplogs(\/\d{4}\/\w{7}/\w{5}\/)starcheck.txt", st
+            r"/data/mpcrit1/mplogs(/\d{4}/\w{7}/\w{5}/)starcheck.txt", st
         ).group(1)
         idcheck = db.fetchone("select id from starcheck_id where dir = '%s'" % mp_dir)
         existing = None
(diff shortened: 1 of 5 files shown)


mica/starcheck/process.py
```diff
@@ -113,73 +104,80 @@ def get_new_starcheck_files(rootdir, mtime=0):
...
-        st = st_with_time['file']
+        st = st_with_time["file"]
         logger.info("Attempting ingest of %s" % st)
         # get an existing id or insert a new one
         mp_dir = re.search(
-            '\/data\/mpcrit1\/mplogs(\/\d{4}\/\w{7}/\w{5}\/)starcheck.txt',
-            st).group(1)
-        idcheck = db.fetchone(
-            "select id from starcheck_id where dir = '%s'" % mp_dir)
+            r"/data/mpcrit1/mplogs(/\d{4}/\w{7}/\w{5}/)starcheck.txt", st
```
jeanconn, line 163: It looks like the previous defect in the regex was at least benign.  Agree that functional testing for the tiny update is sufficient.  This process/ingest module is not tested by the unit tests.

jeanconn (approved):

PR #285, opened 2023-10-04
Set mica agasc calls to use miniagasc*

jeanconn:
## Description

Set mica agasc calls to use miniagasc*

This is in response to https://github.com/sot/agasc/pull/155 which changes the default agasc file to the most recent proseco agasc.  Overall that makes sense for operations, but bits of the mica code expect columns that aren't present in the proseco agasc.  It would probably be useful to update mica to just not use those columns (even for display) but for now I think it makes sense to update the calls in mica to use the miniagasc.  And this may end up being somewhat application specific in mica so I didn't set up any configuration for it.

mica tests don't pass with https://github.com/sot/agasc/pull/155 

## Testing

Tested against a ska3-masters environment with https://github.com/sot/agasc/pull/155 .

### Unit tests

- [x] Linux

- [ ] [PLATFORM]:

### Functional tests

No functional testing.

### DIFF
--- a/mica/report/report.py
+++ b/mica/report/report.py
@@ -446,7 +446,7 @@ def catalog_info(starcheck_cat, acqs=None, trak=None, vv=None):
 
 def get_star(id):
     try:
-        agasc_info = agasc.get_star(id, agasc_file=agasc.get_agasc_filename('miniagasc_*'))
+        agasc_info = agasc.get_star(id, agasc_file='miniagasc_*')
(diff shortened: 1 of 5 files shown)


jeanconn:
Retested after [9eb3930](https://github.com/sot/mica/pull/285/commits/9eb3930f74b29f9cd11e7f5fef56b5c5bc8bcdad)

taldcroft (approved):
LGTM

PR #286, opened 2023-10-16
Handle no-starcat observations better in reports

jeanconn:
## Description

Tiny changes to handle no starcat observations better in reports.  For observations that do not have star catalogs (NMM dwells for IR zone especially) handle the missing star catalog with a handled exception instead of an error and don't try to parse the missing date.

Should stop the "Failed report on 44077" type errors.

## Testing

Tested against a linux ska3-masters environment

### Unit tests

- [x] Linux -

Independent check of unit tests by Javier
- [x] Linux

### Functional tests

Updated the report writing unit test to include a recent ER that has no star catalog.  With this PR that passes -- fails against master.

### DIFF
--- a/mica/report/report.py
+++ b/mica/report/report.py
@@ -337,19 +337,20 @@ def obs_links(obsid, sequence=None, plan=None):
         dir_match = re.match(r'/\d{4}/(\w{3}\d{4})/ofls(\w)', mp_dir)
         mp_label = "{}{}".format(dir_match.group(1),
                                  dir_match.group(2).upper())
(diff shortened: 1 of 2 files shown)


javierggt:
In which loads is this relevant?

jeanconn:
For the record -- mica isn't really "load" specific?  I added one obsid that doesn't have a star catalog (high IR zone) to the test reports.  Note that the report writing just works from Linux with Sybase in general.

javierggt (approved):

PR #297, opened 2024-03-22
Handle filling aca image array differently

### LINKED ISSUE #293: DeprecationWarning: NumPy will stop allowing conversion of out-of-bound Python integers to integer arrays
The following code:
```python
import warnings
warnings.simplefilter("always")
from mica.archive import aca_l0
imgs = aca_l0.get_l0_images(467055635, 467055639, slot=7)
```

Produces this warning
```
/Users/javierg/SAO/miniconda3/envs/ska3-flight-2024.0rc8/lib/python3.11/site-packages/numpy/ma/core.py:429: DeprecationWarning: NumPy will stop allowing conversion of out-of-bound Python integers to integer arrays.  The conversion of -9999 to uint8 will fail in the future.
```

The exact point where this happens is https://github.com/sot/mica/blob/4.35.0/mica/archive/aca_l0.py#L237:
```
dat = dat.filled(-9999)
```

where data is an arbitrary array, but in the specific call that causes the warning during testing the type is
```
dtype([('TIME', '>f8'), ('QUALITY', '>i4'), ('INTEG', '>f4'), ('IMGFUNC1', 'u1'), ('IMGSTAT', 'u1'), ('IMGSCALE', '>i2'), ('BGDAVG', '>i2'), ('IMGRAW', '>f4', (64,)), ('IMGSIZE', '>i4')])
```

jeanconn:
## Description

Handle filling aca image array differently

Fixes #293 

### Unit tests

- [x] Linux

```
ska3-jeanconn-fido> git rev-parse HEAD
a1ca210964cbf815a3adce67bcd5323723b6f9c9
ska3-jeanconn-fido> pytest
================================================================= test session starts ==================================================================
platform linux -- Python 3.11.8, pytest-7.4.4, pluggy-1.4.0
rootdir: /proj/sot/ska/jeanproj/git
configfile: pytest.ini
plugins: anyio-4.3.0, timeout-2.2.0
collected 108 items                                                                                                                                    

mica/archive/tests/test_aca_dark_cal.py ..................                                                                                       [ 16%]
mica/archive/tests/test_aca_hdr3.py .                                                                                                            [ 17%]
mica/archive/tests/test_aca_l0.py ...                                                                                                            [ 20%]
mica/archive/tests/test_asp_l1.py .......                                                                                                        [ 26%]
mica/archive/tests/test_cda.py ..............................................                                                                    [ 69%]
mica/archive/tests/test_obspar.py .                                                                                                              [ 70%]
mica/report/tests/test_write_report.py .                                                                                                         [ 71%]
mica/starcheck/tests/test_catalog_fetches.py ...............                                                                                     [ 85%]
mica/stats/tests/test_acq_stats.py ...                                                                                                           [ 87%]
mica/stats/tests/test_guide_stats.py ....                                                                                                        [ 91%]
mica/vv/tests/test_vv.py .........                                                                                                               [100%]

=========================================================== 108 passed in 517.20s (0:08:37)
```

Independent check of unit tests by Javier
- [x] Linux

### Functional tests

No functional testing.

### DIFF
--- a/mica/archive/aca_l0.py
+++ b/mica/archive/aca_l0.py
@@ -221,6 +221,11 @@ def get_l0_images(start, stop, slot, imgsize=None, columns=None):
     if 'IMGCOL0' not in columns:
         columns.append('IMGCOL0')
 
+    # It doesn't make any sense to include IMGRAW in the meta-data columns
+    # of ACAImages so just remove it if the user has requested that.
+    if 'IMGRAW' in columns:
+        columns.remove('IMGRAW')
+
     slot_columns = list(set(columns + ['QUALITY', 'IMGRAW', 'IMGSIZE']))
     dat = get_slot_data(start, stop, slot, imgsize=imgsize, columns=slot_columns)
 
@@ -233,23 +238,17 @@ def get_l0_images(start, stop, slot, imgsize=None, columns=None):
         if temp in dat.dtype.names:
             dat[temp] -= 273.15
 
-    # Masked array col access ~100 times slower than ndarray so convert
-    # but keep track of those fill values
-    fill_vals = {col: dat[col].get_fill_value() for col in dat.dtype.names}
-
-    imgsizes = dat["IMGSIZE"].filled()
-    imgraws = dat["IMGRAW"].filled(-9999)
-    dat = dat[columns].filled()
-    cols = {name: dat[name] for name in columns}
+    dat["IMGRAW"].set_fill_value(-9999)
+    dat = dat[~dat["IMGSIZE"].mask]
+    imgraws = dat["IMGRAW"].filled()
 
     imgs = []
-    for i, row in enumerate(dat):
-        imgraw = imgraws[i].reshape(8, 8)
-        sz = imgsizes[i]
+    for row,  imgraw in zip(dat, imgraws):
+        imgraw = imgraw.reshape(8, 8)
+        sz = row["IMGSIZE"]
         if sz < 8:
             imgraw = imgraw[:sz, :sz]
-
-        meta = {name: col[i] for name, col in cols.items() if col[i] != fill_vals[name]}
+        meta = {name: row[name] for name in columns if not np.any(row.mask[name])}
         imgs.append(ACAImage(imgraw, meta=meta))
 
     return imgs
--- a/mica/archive/aca_l0.py
+++ b/mica/archive/aca_l0.py
@@ -243,7 +243,7 @@ def get_l0_images(start, stop, slot, imgsize=None, columns=None):
     imgraws = dat["IMGRAW"].filled()
 
     imgs = []
-    for row,  imgraw in zip(dat, imgraws):
+    for row, imgraw in zip(dat, imgraws):
         imgraw = imgraw.reshape(8, 8)
         sz = row["IMGSIZE"]
         if sz < 8:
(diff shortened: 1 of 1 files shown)


mica/archive/aca_l0.py
```diff
@@ -234,20 +234,22 @@ def get_l0_images(start, stop, slot, imgsize=None, columns=None):
...
 
+    imgs = []
     for i, row in enumerate(dat):
         imgraw = imgraws[i].reshape(8, 8)
         sz = imgsizes[i]
         if sz < 8:
             imgraw = imgraw[:sz, :sz]
 
-        meta = {name: col[i] for name, col in cols.items() if col[i] != -9999}
+        meta = {name: col[i] for name, col in cols.items() if col[i] != fill_vals[name]}
```
javierggt, line 252: `imgraw.fill_value` is not -9999, but you filled it with -9999, so the condition `if col[i] != fill_vals[name]` will never be true for `IMGRAW`. Will that be a problem? Or IMGRAW is an exception?

It might be better to do
```
dat["IMGRAW"].fill_value = -9999
imgraws = dat["IMGRAW"].filled()
```

jeanconn: I suppose the ambiguity is that I only made this work for the the default columns (IMGRAW isn't in the default columns) and it isn't expected that one would put the 8x8 imgraw data *in* the metadata for the image.  I should probably just remove imgraw from columns if it is in there.

jeanconn: And your idea to update the fill value for IMGRAW is also a good one.  Technically I think I'd need to do that before making that little dictionary I have from column names to fill values to be 100% consistent (but the idea would still be that we don't have to worry about imgraw in the "add meta data to the ACAImage section").

taldcroft: I think you don't need to do any filling at all except for `imgraws`. For the meta values for each row do:
```
names = dat.dtype.names  # outside the loop
for i, row in ...:
    meta = {name: row[name] for name in names if not dat.mask[name][i]}
```
Technically speaking if `dat["IMGSIZE"]` is masked then that entire row is junk and should just be skipped. Using a filled value for that column would lead to undefined behavior.

taldcroft: And probably the whole thing can be nicer as:
```
# Only keep rows with non-masked IMGSIZE
dat = dat[~dat["IMGSIZE"].mask]
for row,  imgraw in zip(dat, imgraws):
    imgraw = imgraw.reshape(8, 8)
    sz = row["IMGSIZE"]
    if ..:
        ...
    meta = {name: row[name] for name in names if not row.mask[name]}
    ...
```

taldcroft: I think that my 2nd suggestion just works in any case and you don't need to spend time digging. It's all going to be fast enough as well.

jeanconn: Your second suggestion doesn't seem to unmask IMGRAW, and I thought the larger point was not to let the masked array end up as the ACAImage still masked?

taldcroft: That was assuming your change `imgraws = dat["IMGRAW"].filled(-9999)`.

jeanconn: OK.  I think I addressed the comments here.

javierggt (approved):
Looks good to me.

I checked the unit test (including `mica/archive/tests/test_aca_l0.py::test_get_l0_images`) and verified that the deprecation warning does not show.

taldcroft (approved):
LGTM!

